PR #14, opened 2016-12-10
Do not create debug.log

emersion:
Installing browserpass system-wide doesn't work: browserpass tries to create debug.log which fails with "permission denied".

To see error messages, you can launch chromium from the console.


dannyvankooten:
Thanks @emersion. Forgot to remove it, sorry about that!

Issue #16, opened 2016-12-11
When installing Chrome extension by dragging, the installed NativeMessagingHost is incorrect


sharms:
If a extension is dragged from the release package to the extensions tab, the ID is different than `.config/google-chrome/NativeMessagingHosts/com.dannyvankooten.browserpass.json` lists under the `allowed_origins` field

maximbaz:
This is an old issue and I don't think it is valid anymore, as I've seen people installing extension from releases (example: https://github.com/dannyvankooten/browserpass/issues/128#issuecomment-329543935), so I'll close it. 

Let me know if this is not solved for you, and I'll reopen.

PR #21, opened 2016-12-22
Separate searching and fetching of entries, remove username in result list

jplitza:
This way, the GPG key doesn't need to be unlocked during the search, only for
fetching the credentials themselves.

Also, this doesn't send all passwords of all results to Firefox, which I was a
bit uncomfortable with, but only the selected one.

### DIFF
--- a/chrome/host.json
+++ b/chrome/host.json
@@ -4,6 +4,6 @@
   "path": "%%replace%%",
   "type": "stdio",
   "allowed_origins": [
-    "chrome-extension://jegbgfamcgeocbfeebacnkociplhmfbk/"
+    "chrome-extension://gkndoddhagggkjfkojkmnimeafhmpeod/"
   ]
 }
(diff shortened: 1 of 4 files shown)


dannyvankooten:
Hey @jplitza,

I like these changes a lot, thank you for taking the time to shoot in a pull request!

chrome/host.json
```diff
@@ -4,6 +4,6 @@
   "path": "%%replace%%",
   "type": "stdio",
   "allowed_origins": [
-    "chrome-extension://jegbgfamcgeocbfeebacnkociplhmfbk/"
+    "chrome-extension://gkndoddhagggkjfkojkmnimeafhmpeod/"
```
jplitza: Wait, why did I change this? I think this shouldn't have been included in the PR, sorry.

dannyvankooten: Yeah I noticed that, I'm working another issue with the extension ID (#16) so no worries, this'll change before publishing the next release. :)

PR #22, opened 2016-12-24
support for user: field

Midek:
My pass gui saves all login fields as 'user:' instead of username so it would be great if you could support that too.


dannyvankooten:
Hey @Midek, thanks for the PR & sure thing, merged! May I ask which pass GUI that is?

PR #33, opened 2017-01-13
Fixed Shebang and location of chrome-host.json and firefox

x1n4u:
there are location-errors for the `host.json` in the `install.sh` and I changed the shebang to `/usr/bin/env bash`, due to the fact that my bash under nixos can be everywhere


dannyvankooten:
Hi @x1n4u,

The location error should not be there if you run the `make release` command or make sure that you run `make static-files` at least once when developing. The install script is mainly for the final release package, which does not contain subdirectories for Chrome and Firefox.

As for the shebang, nice catch! Could you cherry pick that commit and update this PR?

x1n4u:
Sure if you could guide me, cause i have never done this before... I'm a total git noob.
Also i noticed that the make file is incorect it wants to use google-chrome, but when you have chromium installed you are in trouble

PR #35, opened 2017-01-21
Autoinstall the extension

kcolford:
Installing the extension from the chrome store will cause it to be
installed on every chrome browser your account is sync'd with.  Adding
a system policy that auto installs the extension will install it when
browserpass is available on the system but will leave it off of sync'd
devices that don't want it installed.  Browserpass uniquely has an AUR
package and therefore is one of the few google-chrome <--> pass
bridges that can incorporate this feature.

### DIFF
--- a/chrome/browserpass.json
+++ b/chrome/browserpass.json
@@ -1,0 +1,5 @@
+{
+	"ExtensionInstallForcelist": [
+		"jegbgfamcgeocbfeebacnkociplhmfbk;https://clients2.google.com/service/update2/crx"
+	]
+}
--- a/install.sh
+++ b/install.sh
@@ -70,6 +70,8 @@ mkdir -p "$TARGET_DIR"
 if [ "$BROWSER" == "1" ] || [ "$BROWSER" == "2" ]; then
   echo "Installing Chrome / Chromium host config"
   cp "$DIR/chrome-host.json" "$TARGET_DIR/$APP_NAME.json"
+	mkdir -p "$TARGET_DIR"/../policies/managed/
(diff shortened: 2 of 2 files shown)


kcolford:
If you do want to merge this patch, you should also update the PKGBUILD in the AUR.

dannyvankooten:
This looks amazing @kcolford, can't wait to test drive this. Most likely tomorrow! Thanks a ton for submitting this as a PR.

emersion:
I'll update the AUR package once this gets merged :-)

Do you know if there is something similar for Firefox?

kcolford:
All I could find for Firefox was this <
https://developer.mozilla.org/en-US/Add-ons/Installing_extensions> but I
couldn't find a way to autoinstall it with a device policy like I did with
the chrome one (Firefox is less enterprise friendly than Chrome).

emersion:
Maybe installing the extension to `/usr/share/<vendor>/extensions/<appid>/` is a good idea? (Because the extension must be compatible with the binary anyway)

kcolford:
Maybe, although the management policy was the recommended way for chrome.
But we could certainly try that for Firefox. Although I'm curious why I
couldn't find any other apps out there that install this way for Firefox...

On Sun, Jan 22, 2017, 4:43 AM emersion, <notifications@github.com> wrote:

Maybe installing the extension to /usr/share/<vendor>/extensions/<appid>/
is a good idea? (Because the extension must be compatible with the binary
anyway)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dannyvankooten/browserpass/pull/35#issuecomment-274320355>,
or mute the thread

dannyvankooten:
Sorry I haven't gotten around to this yet, it's not forgotten though.

PR #37, opened 2017-01-30
Search by domain.tld

kcolford:
Instead of the default search going by fully.qualified.domain.tld we
can default to domain.tld so that visiting accounts.google.com will
bring up passwords stored under google.com.  This is probably more
natural for most users.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -181,5 +181,5 @@ function fillLoginForm(login) {
 function parseDomainFromUrl(url) {
   var a = document.createElement('a');
   a.href = url;
-  return a.hostname.replace('www.', '');
+  return a.hostname.replace(/[a-z0-9-]+\.[a-z0-9-]+$/, '$&');
 }


dannyvankooten:
Thanks @kcolford, I was literally running into this all the time when logging into Google. Hope to push out a release for this somewhere next week, in the midst of moving between houses right now.

opatut:
I tried this feature and it seems to not work (yes I installed from git).

It seems like the replacement is faulty logic. `$&` replaces the found *part* (not the whole string) with the matched substring. So in the end, the `replace()` call is a noop.

Try this in your favourite JS environment:
```
'ßöö213n€äßd://om_foo.bar.baz.com'.replace(/[a-z0-9-]+\.[a-z0-9-]+$/, '$&')
``` 

Also, this patch probably breaks with TLDs like `.co.uk` or similar. I am working on a fix.

opatut:
See #44

dannyvankooten:
Hey @opatut,

Nice catch but it seems that this regex is not quite there yet. For example, this will enable the use of .co.uk domains but it'll fail on `.co` extensions.

Changing the first part of the extension regex to accept 2-characters instead of 3 seems to solve this, but I haven't tested it extensively yet.

```
/[-\w]+\.(?:[-\w]+\.xn--[-\w]+|[-\w]{2,}|[-\w]+\.[-\w]{2})$/i
```

PR #43, opened 2017-02-25
install.sh: Fix paths to manifests

eqyiel:
`$DIR/chrome-host.json`, `$DIR/chrome-policy.json` and `$DIR/firefox-host.json` don't exist anymore (it looks like they were moved out of the root directory at some point).


eqyiel:
Whoops, nevermind, it looks like users are supposed to run `make static-files` first.

PR #44, opened 2017-02-28
Fix domain.TLD parsing

opatut:
* actually extract the matched domain name
* improve regex to match domains of TLDs that contain a dot (.co.uk)
* this most likely does not match *all* cases, but it is a simple and
  quick improvement that does not need a lot of logic or another
  dependency

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -177,9 +177,13 @@ function fillLoginForm(login) {
   window.close();
 }
 
+// Thanks to Simon Buchan at http://stackoverflow.com/questions/6449340/how-to-get-top-level-domain-base-domain-from-the-url-in-javascript
+var DOMAIN_REGEX = /[-\w]+\.(?:[-\w]+\.xn--[-\w]+|[-\w]{3,}|[-\w]+\.[-\w]{2})$/i;
+
 // parse domain from a URL & strip WWW
 function parseDomainFromUrl(url) {
   var a = document.createElement('a');
   a.href = url;
-  return a.hostname.replace(/[a-z0-9-]+\.[a-z0-9-]+$/, '$&');
+  var m = a.hostname.match(DOMAIN_REGEX);
+  return m && m[0];
 }
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var m = require('mithril');
+var parseDomain = require('parse-domain');
 var app = 'com.dannyvankooten.browserpass';
 var activeTab;
 var searching = false;
@@ -81,9 +82,9 @@ function init(tab) {
   }
 
   activeTab = tab;
-  var domain = parseDomainFromUrl(tab.url);
-  if( domain ) {
-    searchPassword(domain);
+  var parsedDomain = parseDomain(tab.url);
+  if( parsedDomain ) {
+    searchPassword(parsedDomain.domain + '.' + parsedDomain.tld);
   }
 }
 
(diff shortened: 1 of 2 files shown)


dannyvankooten:
Hey @opatut,

Nice catch but it seems that this regex is not quite there yet. For example, this will enable the use of .co.uk domains but it'll fail on `.co` extensions.

Changing the first part of the extension regex to accept 2-characters instead of 3 seems to solve this, but I haven't tested it extensively yet.

```
/[-\w]+\.(?:[-\w]+\.xn--[-\w]+|[-\w]{2,}|[-\w]+\.[-\w]{2})$/i
```

opatut:
Yes, you're right. I'm gonna play around a little more. Good old [regex golf](http://www.alf.nu/RegexGolf) :)

opatut:
You know what, why not pull in [parse-domain](https://github.com/peerigon/parse-domain), get it right once and for all, and stop worrying. It's not like we need to be concerned about bundle size, nor do we need to worry about license since parse-domain is public domain (no pun intended). Updating PR soon.

dannyvankooten:
Thanks @opatut, that works great for me so far. The only issue I have with this now is that it doesn't search for the full domain (incl. subdomain) anymore, so if you have a specific login for a site's subdomain then it won't show up in the list of available logins.

Ideally I think it would be neat if it prioritises a full exact match but then shows top-level domain matches too. I'm going to take a stab at this.

Deadolus:
tldr: this commit does not fully solve the problem and improves things only for chrome: we should do things in the browserpass executable. I do not have the fix either

Hi all, 

sorry, I have not seen that someone already tried to improve the matching of domains.
As Danny wrote in my pull request #58 , I tried to do something similar but in the executable coming with browserpass. 
I searched for what other password managers do: they seem to have "struggled" with subdomains too,
see for example at [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=385619). 
I have think that the expected behaviour from users is to handle multiple subdomains. 
Maybe offer logins from other subdomains.
My solution for handling multiple subdomains in pass was to add wildcards, e.g. instead of tld.com I have a file with *.tld.com, I thought that was "normal", but it seems like at the moment I can't find any source or anyone other doing things like this. 
So I am left wondering if I actually came up with this myself and may only be an island.
**Ok, so how should, in my opinion, the project fix this?**
I propose to do the searching/filtering in the executable, keeping the browser-dependent code as small as possible. It seems that this fix will only solve things for chrome so far, I think this (or any other filtering problems) still are not solved in the other browsers.
The browser shall send the whole domain name to the executable, which will then parse (sub-)domain and give relevant matches back to the browser.
If we/you don't do it that way we'll have to add filter code to every browser, instead of just one place.
Furthermore, I think my *.tld.com approach/standard may not be optimal and I propose instead: 

- if we are on a.tld.com and pass has a password for "a.tld.com": URL should match
- if we are on a.tld.com and pass has a password for "tld.com": URL should match
- if we are on a.tld.com and pass has a password for "a.tld.com" and "tld.com": URL "a.tld.com" should match
- if we are on b.tld.com and pass has a password for "a.tld.com": URL should NOT match, but maybe suggest somehow to use the password of the other subdomain. 
If we/you fill the password in this case automatically I am a bit concerned about being vulnerable to someone targeting and harvesting users for their passwords on sites where you can create your own "personal page" through subdomains.

My code does not yet handle all tld and subdomains cases correctly (did not think of the .co.uk case, will need to add something similar to parse-domain). 
It also uses the my own suboptimal password scheme (which I think I'll abandon) so far. 
It may probably only be used as a guide where to add the necessary code.

What do you guys think? Any suggestions? Have I forgotten, misunderstood or overlooked something?
I do not have time today, but if no one is faster I might take a stab at it tomorrow.

Cheers, 

Deadolus

mikkeloscar:
@Deadolus great writeup!

I agree with most of what you are saying and have had similar ideas.

One thing which hasn't been mentioned afaik is the functionality of the search field vs. the list of suggestions based on current domain. In my mind the functionality is different and it doesn't make sense to try and solve both cases with the same logic. I would propose to split it up into e.g. `search` and `lookup`:

* `search` logic for the search field, can do fuzzy searching or whatever makes sense
* `lookup` logic for listing relevant pass entries based on the current domain. Like @Deadolus mentioned, this should only show suggestions that are direct matches or matching a subdomain.

I did a quick PoC some weeks ago but never got around to writing down a proposal. You can see the code here: https://github.com/mikkeloscar/browserpass/commit/5b26cbec42cd1453a8f1945a3a15c2eda27d0b18

It's far from perfect, but maybe it can serve as inspiration for this discussion.

satyakid:
@Deadolus,

I seem to have ended up implementing your idea, though I came up with it independently :-). Basically search matches the most specific domain. satyakid@51579bc12

dannyvankooten:
Hey everyone,

Thanks for chiming in here. I like the idea of keeping "search" and "lookup" separated so that search queries can be as specific and relevant as needed. 

The PR submitted by @satyakid looks promising as well, I hope to be squeeze in a few hours this week so that we can move forward on this & push out a release. Thanks so far for the great work & thoughts!

PR #45, opened 2017-03-01
Refactoring Go code

emersion:
- Adds a `pass` package with a `Store` interface - will allow new stores implementation (e.g. a Git store, see https://github.com/emersion/webpass/blob/master/config/auth_git.go#L122)
- Use streaming APIs instead of `[]byte`
- Use a `cmd/browserpass` package for `main` instead of putting it in the repo's root and not being able to split the code in multiple files
- Better testability (I just ported the old tests, but it should be easier to ad new ones)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
-browserpass
-browserpass-*
+/browserpass
+/browserpass-*
 *.crx
 *.log
 *.pem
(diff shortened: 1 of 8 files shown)


dannyvankooten:
Looks great, thanks a ton @emersion!

PR #50, opened 2017-03-13
Load json into struct

mikkeloscar:
This is just a minor style issue I found looking at the code. It's cleaner to load the json data into a struct rather than a map. :)


dannyvankooten:
Definitely, much cleaner! Thanks @mikkeloscar!

PR #52, opened 2017-03-17
fix broken install on OSX

stensonb:


dannyvankooten:
Hah, don't worry @stensonb, you're not the first to run into this so probably it is the project that should be a little more clear about this!

PR #55, opened 2017-03-20
#46 put password in all frames which contain visible password field

### LINKED ISSUE #46: Uncaught TypeError - Not working with login box inside embedded frame.
Browser pass is not working with twitch.tv, returning a `Uncaught TypeError: Cannot read property 'form' of null`. Looking at the Chrome Developer window, it is extracting the password from Pass. I suspect this has something todo with Twitch not having the login form on the page, instead loading a frame of https://passport.twitch.tv/authentications/.

maximbaz:
This is a pretty naive approach to fix #46, where instead of pasting password in the top frame, the extension will try paste it in all available frames (this requires adding more permissions). 

To do as little as possible in the frames which don't have the password field, I added an early return. While testing I also discovered that some websites have frames with _hidden_ password fields - if user doesn't see them, these we want to exclude as well.

Like it was before, if a page has two visible password fields, we just pick the first one.

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -12,7 +12,9 @@
   },
   "permissions": [
     "activeTab",
-    "nativeMessaging"
+    "nativeMessaging",
+    "http://*/*",
+    "https://*/*"
   ],
   "commands": {
    "_execute_browser_action": {
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -149,12 +149,23 @@ function fillLoginForm(login) {
 
   var code = `
   (function(d) {
+    function queryFirstVisible(parent, selector) {
+      var elems = parent.querySelectorAll(selector);
+      for (var i=0; i < elems.length; i++) {
+        var style = window.getComputedStyle(elems[i]);
+        if (style.visibility !== 'hidden' && style.display !== 'none') {
+          return elems[i];
+        }
+      }
+    }
+
     function form() {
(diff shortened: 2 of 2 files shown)


dannyvankooten:
Thanks a ton for putting this together @maximbaz. I haven't yet had time to look into this in detail but hope to be able to this Friday. 

From first glance, I'd prefer not to grant the browser add-on permissions to access any and all websites so I was hoping this could be fixed by traversing into the iframe content  using `iframe.contentDocument`. Not too sure whether's that's actually possible though.

Either way, thanks a lot for taking the time to look into this and putting this PR together!

maximbaz:
I was pretty frustrated as well with giving these permissions, but unfortunately this is the only way. The regular javascript will not be allowed to access inner frames (not sure if this is by default or inner frame configures this, but still):

image

maximbaz:
Hey @dannyvankooten, I wonder if you had a chance to look at this yet? I've been using this code ever since I created this PR, and so far haven't seen any issues. But I'd like to switch back to the official extension once this is merged 😉

irgendwie:
Works like a charm for a website I had the same problem with. :+1:

dannyvankooten:
@maximbaz not yet, but this Friday is my next "funday" so will then hopefully have some time for it!

maximbaz:
A friendly ping @dannyvankooten 🙂

dannyvankooten:
Thanks @maximbaz, finally got around to test driving it. I also needed some time to be at peace with the extended permissions for the browser add-on. :smile:

PR #56, opened 2017-03-21
Add instructions for building the Linux binary with Vagrant

derpston:
I found the dependency notes in CONTRIBUTING a little light and I'd like to encourage others to get involved, so I put together some instructions for easily and repeatably getting a build environment set up for browserpass.

This is intentionally incomplete, it doesn't address installing chrome or doing a full build, I'm intentionally trying to just get this one thing done and hoping it's useful enough to others already.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,68 @@ You will need Node, [Yarn](https://yarnpkg.com/) and Golang installed.
 
 The command above will generate packed extensions for both Firefox and Chrome and compile the Go binaries for Linux and MacOSX.
 
+## Setting up a Vagrant build environment and building the Linux binary
+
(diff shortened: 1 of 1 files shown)


felixb:
would you mind adding the Vagrantfile as well?

derpston:
I.. could do. I originally didn't as is a completely generic one written by `vagrant init` and I tend to shy away from committing anything machine-generated because the value is generally very low, and it just becomes a source of confusion later.

If you're still keen on having a Vagrantfile I'll throw it in - what would you like to do?

felixb:
Well, I'll create my own Vagrantfile to test things. Thanks anyway.

derpston:
Oh, maybe I'm misunderstanding you then. When I did this, I ran `vagrant init minimal/xenial64` and it produced a Vagrantfile with everything necessary, no need to create one manually.

Is there a case where one would need to create one manually, or to edit the one generated? Wondering what I'm missing here. :)

felixb:
You'd want a Vagrantfile which provisions the VM automatically as you described in the CONTRIBUTING.md. The user should do nothing but call `vagrant up`.

maximbaz:
I'll close this in favor of #106 🙂 Thanks for the contribution!

PR #58, opened 2017-04-04
Wildcard passwords

Deadolus:
These changes will enable wildcard password files support. 
For example *.google.com will match "accounts.google.com", "photos.google.com", and so on...
Only one level of subdomains is supported so far: "photos.accounts.google.com" would not be matched.

### DIFF
--- a/pass/disk.go
+++ b/pass/disk.go
@@ -46,7 +46,16 @@ func (s *diskStore) Search(query string) ([]string, error) {
 		return nil, err
 	}
 
-	items := append(matches, matches2...)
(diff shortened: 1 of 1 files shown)


dannyvankooten:
Hey @Deadolus,

Thank you for submitting a PR, this looks neat! #44 does something similar though, it will only send the top-level domain from the browser extension to the Go binary. Does this PR conflict with that or can these co-exist? I'm not too sure by skimming through.

Thanks a lot anyway!

Deadolus:
Closing this as it does not fully solve the problem, see discussion in #44

PR #61, opened 2017-04-09
Nicer domain search.

satyakid:
If domain doesn't match, then strip off the first component and
search again. Continue till we run out of domain components or
find a match.

For example if foo.bar.baz.box.com is searched for then this leads
to the following searches (in order):
  foo.bar.baz.box.com
  bar.baz.box.com
  baz.box.com
  box.com
The search terminates at the first one which yields a non-empty
result.

### DIFF
--- a/pass/disk.go
+++ b/pass/disk.go
@@ -34,14 +34,30 @@ func defaultStorePath() (string, error) {
 }
 
 func (s *diskStore) Search(query string) ([]string, error) {
+	// Number of prefix matches to be checked is the number of periods
+	// with lower bound of 1
+	count := strings.Count(query, ".")
+	if count < 1 {
+		count = 1
+	}
+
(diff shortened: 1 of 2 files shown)


mikkeloscar:
It seems quite inefficient to read from the disk so many times for a search. It would be better to read once, and then search once the filenames are in memory.

satyakid:
@mikkeloscar: I modified the code to do the search once as you suggested.

Thor77:
I really like this approach, just tested with Firefox Nightly (55.0a1) on ArchLinux, works fine!

maximbaz:
I believe this has been handled by #84, and from my experience subdomain search is working fine now. Since this is a quite old PR with conflicts, I'm going to close it to clean things up - please open a new PR if there is something you still want to add.

PR #62, opened 2017-04-09
Implement lookup separate from search

mikkeloscar:
As discussed in #44 I think it makes sense to split the behavior of the search field and the list of domains shown in browserpass.

This is my PoC for separating search and lookup.

The algorithm used for lookup is a follows:

Assuming you have the following domains in your store:

```
my.sub1.domain.tld
sub1.domain.tld
sub2.domain.tld
domain.tld
sub1.domain2.tld
domain2.tld
```

and you are on page `sub1.domain.tld`, then the lookup will return the following list of domains `[sub1.domain.tld, domain.tld]` ordered by most precise match.

As this is a PoC not everything is implemented yet:

* [ ] handle domains stored as `domain.tld.gpg` (currently only handles `domain.tld/user.gpg`).
* [ ] handle multilevel tlds correctly e.g. `co.uk`.
* [ ] consider stripping `www.` automatically.

I also haven't benchmarked the implementation, but I suspect it's on par with the current zglob implementation.

I realize this is somewhat conflicting with #61 and I'm fine with you going that way if you prefer.

### DIFF
--- a/browserpass.go
+++ b/browserpass.go
@@ -56,6 +56,12 @@ func Run(stdin io.Reader, stdout io.Writer, s pass.Store) error {
 				return err
 			}
 			resp = list
+		case "lookup":
+			list, err := s.Lookup(data.Domain)
+			if err != nil {
+				return err
+			}
+			resp = list
 		case "get":
 			rc, err := s.Open(data.Entry)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


maximbaz:
I believe this has been handled by #84, and from my experience subdomain search is working fine now. Since this is a quite old PR with conflicts, I'm going to close it to clean things up - please open a new PR if there is something you still want to add.

PR #64, opened 2017-04-12
Fix 'type=email' not available by using 'name=email'.

irgendwie:
Some websites, e.g. Netflix (https://www.netflix.com/de-en/login) doesn't use `type='email'` or `type='text'` for the username field. Instead the field can/must be identified by using `name='email'`.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -173,7 +173,7 @@ function fillLoginForm(login) {
     }
 
     update(field('input[type=password]'), ${JSON.stringify(login.p)});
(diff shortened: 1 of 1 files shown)


Scrunshes:
:+1: I just observed the same problem and thought of the same solution, would be really cool If this PR could be merged :-)

dannyvankooten:
Hey @irgendwie,

Nice catch and thanks for submitting yet another PR! I didn't quite like adding name-attribute specific query selectors as this is probably bound to happen again so I went with a more general approach by falling back to the first input element in the password form, which I guess should suffice in most cases.

PR #66, opened 2017-04-12
form() now only returns the first visible element.

irgendwie:
When calling `form()`, the first form is given also if it isn't visible for the user (hidden forms).
With this PR, all forms which contain a `input[type=password]` are searched - again returning the first element - but only if it is visible.

For test, e.g. https://twitter.com/

Edit: https://github.com/dannyvankooten/browserpass/pull/55 already contains a better patch for the same problem.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -150,7 +150,10 @@ function fillLoginForm(login) {
   var code = `
   (function(d) {
     function form() {
-      return d.querySelector('input[type=password]').form || document.createElement('form');
+      var elems = d.querySelectorAll('input[type=password]');
+      for (var el of elems)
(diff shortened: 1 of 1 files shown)


dannyvankooten:
Thanks @irgendwie, I went with #55 but appreciate you submitting a PR. I like checking for the parent offset to determine whether a field is visible, would it make sense to modify #55 to do that as well?

maximbaz:
With the current check I tried to catch not only invisible fields, but also very narrow visible fields (I saw this somewhere, I assume it was a spam check, when the field had any value the request was rejected). 

Just a consideration, perhaps see if you can find an example of when the current check is not working as good as checking for the offsetParent before adding it.

PR #68, opened 2017-04-28
add support for OpenBSD

qbit:
- Add support for OpenBSD
- Checks for common names of Chrome and uses the one it finds.


dannyvankooten:
Looks great. Thanks @qbit!

PR #78, opened 2017-05-14
Completion for password only input

Scrunshes:
With the actual behavior if the page only has a password input (password confirmation page...) then it will be filled and overridden by the login afterwards because it's the first of type input. If we change the order than it will correctly work.
I added a comment so that this behavior will not be lost overtime which can easily happened with this kind of details.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -200,8 +200,9 @@ function fillLoginForm(login) {
       return;
     }
 
-    update(field('input[type=password]'), ${JSON.stringify(login.p)});
+    // Order must be : login than password, so that password only page can be completed
     update(field('input[type=email], input[type=text], input:first-of-type'), ${JSON.stringify(login.u)});
+    update(field('input[type=password]'), ${JSON.stringify(login.p)});
 
     var password_inputs = queryAllVisible(form(), 'input[type=password]');
     if (password_inputs.length > 1) {


maximbaz:
I feel the problem is not the ordering, but that the username field is searched with (among others) as `input:first-of-type'`, so the first input field in the form is always considered as a username, even if it is marked as `[type=password]`. @dannyvankooten do you know what was the reason behind that?

If we just drop the `input:first-of-type'` when searching for username, order will not matter anymore.

Filling only empty fields is an interesting idea, but I'm not 100% sold on that yet... What if rogue Chrome decides to auto-populate it, or something... 😄

maximbaz:
Oh, it appears to be a pretty recent addition actually, so I'm sure you still remember it 😄 

I actually believe the originally proposed fix in #64 was a better idea, exactly for the reason brought up by @Scrunshes: you cannot assume that the first input is always a username field. We better keep detecting username field with some specific selectors.

Scrunshes:
I agree with @maximbaz, for the moment it's a good fix.

My 5 cents: it might be a good idea to add a feature to let user specify which fields are what on a site but it has some loopholes : it would make things way more complicated but on the other hand with a bit of intelligence we could use that data to improve the selectors for everybody...

maximbaz:
Would you update your PR to revert what you did and instead replace `input:first-of-type` with `input[name=email]` as was initially suggested in #64? If @dannyvankooten agrees, it would be easier for him to merge.

Your idea is interesting, but I can't think of an easy way to implement it 🙂 I guess people creating github issues with examples is the simplest approach we have today...

Scrunshes:
@maximbaz I'm totally okay with adding that although I'm not sure about removing input first. I totally agree that's a poor man's solution to the problem but It might be a real fix for many sites (hard to quantify anything here).  Let's see what other people think...

Scrunshes:
What's the status here ? Do we merge ? Do we agree on another solution ? Was it solved in an other PR ?

edouardmenayde:
Closing the PR in favor of #101 which uses a smarter approach.

PR #82, opened 2017-05-21
Added subdomain part to domain search, fixes #81

### LINKED ISSUE #81: Unable to find passwords on URLs with sub-domain
Expected:

When I'm on the page "manage.statuspage.io", the extension should find my password stored at `manage.statuspage.io/...`

Actual:

The extension appears to search for "statuspage.io" (which does work with `pass find`), and finds no results.

Steps to reproduce:

1. Store a password for `subdomain.something.com/username`
1. Visit `subdomain.something.com`
1. Use the extension

Details:

browserpass

This has been happening since the most recent update, the one that requested more permissions. Before then, this was working great. Passwords for URLs without a sub-domain (e.g. `github.com/pbrisbin`) are working fine.

I didn't fully understand the logic of when the extension ignored sub-domains or not, but however it worked previously felt right and whatever it's doing now has drastically increased the miss rate for me. As a naive developer I would expect the extension to delegate fully to `pass find {current url}`, which works consistently.

Miterion:
With the parse-domain library the domain gets split up into three parts, subdomain, domain, tld. The implementation combined the domain and tld part but ommitted the subdomain part. Therefore a search for `test.example.com` would only return domains containing `example.com` and omitting everything else.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -87,7 +87,7 @@ function init(tab) {
   });
 
   if( parsedDomain ) {
-    var searchDomain = [parsedDomain.domain, parsedDomain.tld]
+    var searchDomain = [parsedDomain.subdomain, parsedDomain.domain, parsedDomain.tld]
       .filter(function (x) { return x; })
       .join('.');
 


varac:
@dannyvankooten Could you do a relase based on this PR pls ? 80% of my logins doesn't work anymore for me (and I guess roughly the same amount for everybody else), so I consider this a severe bug that should be covered by a hotfix release.

PR #101, opened 2017-07-03
guess username field by id

felixb:
Guess username fields by id instead of just picking the first one.
The current behavior overrides the account field for the aws console.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -195,14 +195,19 @@ function fillLoginForm(login) {
 	}
 
 	update(field('input[type=password]'), ${JSON.stringify(login.p)});
-	update(field('input[type=email], input[type=text], input:first-of-type'), ${JSON.stringify(login.u)});
+	// Guess username field
+	var usernameFields = field('input[id=username], input[id=user_name], input[id=userid], input[id=user_id], input[id=login], input[id=email]');
+	if (! usernameFields.id) {
+	  usernameFields = field('input[type=email], input[type=text]');
+	}
+	update(usernameFields, ${JSON.stringify(login.u)});
 
 	var password_inputs = queryAllVisible(form(), 'input[type=password]');
 	if (password_inputs.length > 1) {
 	  password_inputs[1].select();
 	} else {
 		window.requestAnimationFrame(function() {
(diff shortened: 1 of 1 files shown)


maximbaz:
Related: https://github.com/dannyvankooten/browserpass/pull/78#issuecomment-301589526

Guessing by id is a good idea, but I think that "first of type" should be removed entirely.

chrome/script.browserify.js
```diff
@@ -195,14 +195,19 @@ function fillLoginForm(login) {
 	}
 
 	update(field('input[type=password]'), ${JSON.stringify(login.p)});
-	update(field('input[type=email], input[type=text], input:first-of-type'), ${JSON.stringify(login.u)});
+	// Guess username field
+	var usernameFields = field('input[id=username], input[id=login], input[id=email]');
```
maximbaz: why not just pass all arguments here at once?

```
var usernameFields = field('input[id=username], input[id=login], input[id=email], input[type=email], input[type=text]');
```

felixb: you'd get the first match, in case of the AWS console, it's still the account field.

maximbaz: could you explain that a little? in my example I excluded the `first-of-type` and rearranged fields, so I think the two code snippets will be identical... or not?

felixb:
In my experience, rearranging the filters did not change the order of the outcome. I guess the list of matches is ordered by appearance in the DOM. Maybe I'm mistaken though.

felixb:
I just removed the `input:first-of-type` and fixed the fallback when no `username` field is found.

felixb:
do you need anything from me to merge this change?

maximbaz:
@dannyvankooten ^

felixb:
@dannyvankooten: do you have any opinion on this?

BTW: This fixes #111.

maximbaz:
I rebased and merged your change in https://github.com/dannyvankooten/browserpass/commit/adaa3a7faefc4b4b79e02c4b84c57a610c1c0394

You were right that the order or selectors did not matter, I fixed that separately and it allowed me to simplify your change. Let me know if it works for you - and thanks!

felixb:
Works great!

PR #106, opened 2017-07-10
Contributing vagrant build

felixb:
This is a simplified version of the Vagrant build environment proposed in #56.


maximbaz:
There is no reason not to merge this PR, thanks for the contribution 🙂

PR #109, opened 2017-07-14
Add support for 64-bit FreeBSD

cstroe:
The compatibility list on addons.mozilla.org should be updated:

browserpass-on-freebsd

Installing the plugin for Linux works just fine.


maximbaz:
Thanks, merged in https://github.com/dannyvankooten/browserpass/commit/8723d9cf3ce835e8e1e0e169d57b5f5d6f8be116

@dannyvankooten please update the compatibility list next time you upload an extension, as suggested above.

PR #114, opened 2017-07-27
Update install.sh to find the right files

thiderman:
It seems that the file paths for the manifest and host files have changed and `install.sh` was not updated to match this.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -92,11 +92,11 @@ mkdir -p "$TARGET_DIR"
 
 # Copy manifest host config file
 if [ "$BROWSER" == "1" ] || [ "$BROWSER" == "2" ] || [ "$BROWSER" == "4" ]; then
-  cp "$DIR/chrome-host.json" "$TARGET_DIR/$APP_NAME.json"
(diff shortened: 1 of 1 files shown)


maximbaz:
Hey @thiderman, I think the reason `install.sh` is using `chrome-host.json` and others is because it is expecting that `make static-files` has been run on the repo. However, since `make static-files` is only doing files copy and nothing else, I don't really see the benefit of it, and I like your approach better.

Could you therefore also cleanup the `static-files` goal from `makefile`?

maximbaz:
I ended up doing this myself as I was editing that file anyway - thanks for the contribution!

maximbaz:
FYI the paths were correct and I ended up reverting the change in https://github.com/dannyvankooten/browserpass/commit/45374cd95e632a1099fe2d1af9f5f5ebf9896e41 as it broke release archives. 

The missing puzzle piece was that when you run `./install.sh` from source code, you have to go through the build steps described in [CONTRIBUTING.md](https://github.com/dannyvankooten/browserpass/blob/master/CONTRIBUTING.md). Specifically, the step `make release` will ensure the presence of all the files needed by `install.sh`.

Issue #104, opened 2017-07-07
browserpass && GnuPG-card needs a second click after entering the PIN


gurucubano:
I'm using behind browserpass the GnuPG-card and to allow access to the secretkey, one must enter a 6 digit PIN. The card then remains unlocked until withdraw. When I need some web credentials in the first time after card insert, the click on the browserpass icon let me select the correct keyfile from password-storage and the request results in a PIN-entry pop-up which unlocks the secret fine. But, the credentials are not written into the browser fields. It need a second click on the icon to get them now written correctly.

Is there anything I could debug to help?

Matthias

maximbaz:
The issue is more generic, and it's a duplicate of #83 (also #86 and #87)

maximbaz:
This has been fixed in `master` branch, thus closing for cleanup purposes.

Issue #87, opened 2017-06-05
Form not filled when gpg asks for a password


jayme-github:
OS Linux
pass 1.7.1
browserpass 1.0.4
chrome 58

If gpg prompts for my passphrase when selecting a password from browserpass, the username/password is not filled after entering the gpg passphrase.
I need to re-select the password in browserpass to have the login form filled correctly (with gpg passphrase in gpg-agent's cache).

jayme-github:
Autofill always works when inspect popup is open (#86)

maximbaz:
Also possibly related: #83

jayme-github:
Closing in favor of #83 as it has more history and details.

Issue #86, opened 2017-06-04
Only autofills in after clicking 'Inspect Object' in developer mode


lobachevzky:
The title says it all. I made a quick screencast and posted on youtube so you can see what I'm talking about:
https://www.youtube.com/watch?v=ToXt9nHovO8&feature=em-upload_owner

jayme-github:
This is a duplicate of #83 (#87).

lobachevzky:
Can you explain what those issues have to do with clicking 'Inspect Object'?

maximbaz:
When pinentry dialog opens, the browserpass popup gets closed. When it is closed, it is destroyed - it will not receive your password and will not fill it on the website. This is how chrome extension popups are designed, and there is nothing we can do about it.

However, when you "Inspect popup", the developer panel keeps the popup alive, and so even though pientry dialog was opened, the popup is still there and it still receives the password and is able to fill it on the page.

See #83, there were some links posted that explain this behavior.

lobachevzky:
I see. Thank you.

Issue #83, opened 2017-05-26
I have to press twice


joyarzun:
In the first time gpg ask me for a password, okay I enter it but nothing happened. Then I have to press browserpass again to fill the form.

Browserpass version: Master branch (1.0.4)
OS: MacOs
Browser: Opera last and Chrome 58

maximbaz:
Yes, I see this as well. The first time the key is being added to gpg-agent but after action stops, the second time the key is already in the agent so password is successfully retrieved.

raxod502:
This issue happens for me because my pinentry program (pinentry-mac) creates and focuses a new window, which causes the browserpass dropdown menu to lose focus and disappear. This could be resolved by making the dropdown menu less eager to disappear as soon as it loses focus.

joyarzun:
Do you know why this happen with mac pinentry and do not with others like linux pinentry? For example, I work with Manjaro Gnome flavor and the issue does not occur. Anyway, do you figure out how to work around?

raxod502:
Presumably it's a difference in how the window manager handles window focusing. I can't test on macOS, since all other pinentry programs are horribly broken for me. There is no workaround that I can see other than just entering my password twice (since gpg-agent will cache it).

Aside from it causing this issue, I think the fragility of the popup is a bug in its own right, since I frequently am interrupted in the middle of an operation and don't want whatever I'm working on to disappear on me.

jayme-github:
Happens for me too:
OS Linux
pass 1.7.1
browserpass 1.0.6
chrome 59

Seems like i'ts not an easy task to prevent the addon from closing when using focus (https://stackoverflow.com/questions/4146930/how-to-keep-google-chrome-extension-popup-open). 
Maybe it is possible for the addon to complete it's work even when closed? (Must admit that I have almost no idea about addon development for both chrome and firefox).

joyarzun:
@jayme-github It doesn't happen to me in Linux. Can you change your pinentry?

jayme-github:
@joyarzun I probably can - which do you use?
Mine is pinentry-gnome3 currently... (running a KDE5 environment this seems pretty odd :smile: )

mlen:
Happens for me with `pinentry-gtk-2`

mlen:
Another thing is that this may depend on the window manager being used, for me it is awesomewm 4.0

jayme-github:
[...]

PR #119, opened 2017-08-16
Use background page for filling in forms

irtimmer:
This pull request fixes the issue of Browspass canceling the filling in when the popup loose focus.
By using a background page Browserpass can continue to retrieve the password even when the popup is closed.
This therefore should fix the issues #104, #83, #86 and #87


jayme-github:
Great!
This works like a charm for me. Plus: My guess was right :smile: 

Thank you very much!

maximbaz:
Nicely done, works great!

PR #139, opened 2017-09-18
Select the fields without consideration of the case

jannickfahlbusch:
The case of the fields name shouldn't be considered while selecting the the fields for the username/password.

[Here](https://drafts.csswg.org/selectors-4/#attribute-case) is the spec for the _i_ identifier.


chrboe:
Is this actually widely supported? It seems like the document you linked is in a pretty early state. (If both Chrome and FF support it back to a reasonably old version it's fine imo even if it's a draft)

maximbaz:
[Support is good](http://caniuse.com/#feat=css-case-insensitive)

image

Thanks @jannickfahlbusch!

jannickfahlbusch:
You're welcome! :smile:

PR #140, opened 2017-09-18
add pledge(2) on OpenBSD

qbit:
Pledge (OpenBSD only) is a way to restrict the operations an application can make use of on OpenBSD. Applications declare their intent by specifying the system calls they will use, after which point, any use outside of the pledge'd calls will result in the kernel killing the process. This lets us do things like ensure that `browserpass` will never be able to access the network, or look something up via DNS.

More information on pledge can be found [here](http://man.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/pledge.2)


chrboe:
Good idea! Regarding the vendor directory, wouldn't it be possible to utilize git submodules to implement this? (I can't remember off the top of my head, but I think you can give it a specific revision to check out?)

154k lines of third party code is a little much imo 😄

maximbaz:
I already heard [the argument](https://github.com/dannyvankooten/browserpass/pull/73#issuecomment-330100515) for having `vendor` folder directly in our repo, but I'm still not very comfortable with it.

@qbit how about we gitignore the `vendor` folder, but add a new make goal to generate a source archive with all dependencies bundled by running `dep ensure`? Then package managers can use that new release archive.

qbit:
I think that would be win win!

maximbaz:
Cool :) Could you do it, in this PR or in another one? Alternatively I can look into this in the evening.

qbit:
```
qbit@slip[0]:browserpass[openbsd_pledge *]λ find node_modules/ -type f | wc -l    
    2106
qbit@slip[0]:browserpass[openbsd_pledge *]λ find vendor/ -type f | wc -l       
     355
qbit@slip[0]:browserpass[openbsd_pledge *]λ 
```
Feels dirty removing `vendor` but not `node_modules`. I can do it - but I would prefer to do it in another PR.

maximbaz:
`node_modules` is already gitignored 🙂

https://github.com/dannyvankooten/browserpass/blob/05fe33979365c7eb14fbc66298bfe5a9fde08071/.gitignore#L14

Then let's do that PR first, and this PR will become much simpler once you rebase.

maximbaz:
By the way, that's why CONTRIBUTING.md asks to run `yarn` - to fetch the `node_modules` folder. So I guess in the other PR in addition to adding a new make goal and gitignoring `vendor` folder let's also add 
that `dep` tool is needed (besides `Node, Yarn and Golang`) and `dep ensure` after `Run yarn` in the CONTRIBUTING.md

qbit:
Good call - should have checked .gitignore - I will get something put up later today!

qbit:
I have a go question.. I remembered that I can call pledge sooner (via `cmd/browserpass/main.go`), but, in order to do that.. it seems like I need to duplicate the `browserpass_{generic,openbsd}.go` files and make them package with `main`.. is there a better way to go about it?

maximbaz:
Rebased on `master`, tested that it compiles and merged - much cleaner change now, thanks once again!

https://github.com/dannyvankooten/browserpass/commit/b9b9173867626c17cbe90448d526d669def9035d

PR #144, opened 2017-09-19
Add Dockerfile for builds within docker

jayme-github:
Just in case anyone is interested. This can be used to build browserpass (go and js) with docker-powers:
```
# Run once
docker build -t browserpass-dev .

# Repeat while hacking & for updates
docker run --rm -v "$(pwd)":/app browserpass-dev browserpass js
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM golang:1.8-alpine3.6
+
+ENV CGO_ENABLED=0 \
+    APP_PATH="$GOPATH/src/github.com/dannyvankooten/browserpass"
+# New ENV statement as this depends on APP_PATH
+ENV PATH="$PATH:$APP_PATH/node_modules/.bin/"
+
+RUN apk add --no-cache bash git make tar yarn zip && \
+    go get -u github.com/golang/dep/cmd/dep && \
+    mkdir -p $APP_PATH && \
+    ln -s $APP_PATH /
+
+WORKDIR $APP_PATH
+
+ENTRYPOINT ["make"]
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ chrome/inject.js
 node_modules
 .vagrant/
 vendor/
+.*.swp
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # CONTRIBUTING
 
-You will need Node, [Yarn](https://yarnpkg.com/), Golang and [dep](https://github.com/golang/dep) installed.
+You will need docker or Node, [Yarn](https://yarnpkg.com/), Golang and [dep](https://github.com/golang/dep) installed.
 
 ## To build
 - Open `makefile` and, if needed, change `google-chrome` to the appropriate name of your Google Chrome executable (in Linux, it could be google-chrome-stable)
@@ -8,6 +8,22 @@ You will need Node, [Yarn](https://yarnpkg.com/), Golang and [dep](https://githu
 - Run `make release`
 
 The command above will generate packed extensions for both Firefox and Chrome and compile the Go binaries for Linux and MacOSX.
+
+## Build using Docker
+
+The [Dockerfile](Dockerfile) will set up a docker image suitable for running basic make targets such as building frontend and backend.
+
+The `chrome` target is not supported by now (therefore `release` target will not work).
+
+To build the docker image run the following command in project root:
+```shell
+docker build -t browserpass-dev .
+```
+
+To build browserpass (frontend and backend) via docker, run the following from project root:
+```shell
+docker run --rm -v "$(pwd)":/browserpass browserpass-dev browserpass js
+```
 
 ## Setting up a Vagrant build environment and building the Linux binary
 


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.8-alpine3.6
+
+ENV CGO_ENABLED=0
+
+RUN apk add --no-cache bash make zip chromium nodejs-npm && \
+    npm install -g browserify && \
```
maximbaz: this better just use `yarn` instead of `npm install ...`, it will fetch all needed dependencies of the specific fixed versions.

maximbaz:
I love it :+1: Could you also add a note to CONTRIBUTING.md about the possibility to use Docker? Perhaps a section above Vagrant, since Docker is more trending these days 😛

Dockerfile
jayme-github: Im not into this js things very much...just installing and running `yarn` does not seem to work as I'm still missing `browserify` afterwards.

maximbaz: does it produce any output? the difference is that `npm install` will install the very latest version, while `yarn` will install the specific version that I tested as working.

try `yarn` or `yarn install`, but I think it is the same command.

jayme-github: I see, `yarn` installs binary's to `./node_modules/.bin` rather that a "global" path.

jayme-github:
Will happily add notes to `CONTRIBUTING.md` tomorrow if you're okay with the Dockerfile now.

maximbaz:
I'll test the Dockerfile later today, but from the first glance seems very good!

maximbaz:
I tested it a little bit, the `js` goal works nicely, you definitely handled the `yarn` stuff :+1: The `clean` goal doesn't work only because `git` package is missing - just add it, and everything will be fine. The `chrome` goal requires me to add a `chromium-browser` as a potential browser name (will do now on master, make sure to rebase to get this) but still crashes with the following error:

```
"/usr/bin/chromium-browser" --disable-gpu --pack-extension=./chrome --pack-extension-key=./chrome-browserpass.pem
Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
[13:13:0919/182534.019643:FATAL:zygote_host_impl_linux.cc(182)] Check failed: ReceiveFixedMessage(fds[0], kZygoteBootMessage, sizeof(kZygoteBootMessage), &boot_pid).
make: *** [makefile:12: chrome] Aborted (core dumped)
```

Let me know if it works on your end. If you don't want to support the `chrome` goal (and thus its derivatives like `release` goal), I guess you can remove the `chromium` from being installed altogether in the Dockerfile. 

And @qbit will be (1) gitignoring `vendor` folder and (2) adding a yet another make goal to archive sources today or tomorrow. Supporting the new make goal is probably not really needed, but the first point would definitely require changes to Dockerfile (to install [dep](https://github.com/golang/dep) along with `yarn`), so one of you guys would need to do it.

jayme-github:
Thanks for the feedback! I tried around a bit with chromium but I'm still unable to make it build the extension. At least it needs `apk add eudev-libs ttf-freefont` and the extra options `--headless --no-sandbox` to not dump core and/or exit with errors but I still don't get a `chrome.crx` :-/

maximbaz:
I'm thinking maybe if this gets complicated, don't spend time on making chromium work, because for most people it is enough to run `make clean js browserpass-linux64` to test some feature, especially since both Chromium and Firefox allow loading "unpacked" extension, which is the result of `make js`. Compiling `crx` file is only needed for me to publish a release, and I'll not use docker for it.

maximbaz:
Since #148 is merged, there is a new goal `make deps`, all it does is it runs `yarn && dep ensure`, and to make it work the `dep` binary is needed now. 

I installed it on my laptop with the following command: `$ go get github.com/golang/dep/...`

There is also a new goal called `make tarball`, which I didn't test on this Dockerfile, but I assume it should work as long as `tar` is available on the system, so you might need to add it to be installed in the Dockerfile.

It would be nice if you can make chromium work, but if not, don't worry, as I said, I personally see the goals `make clean deps js static-files browserpass-***` as the most important, that is what a regular user would need, the rest is optional.

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENV CGO_ENABLED=0 \
+    PATH="$PATH:/app/node_modules/.bin/"
+
+RUN apk add --no-cache bash make zip chromium yarn && \
+    yarn install && \
+    mkdir -p $GOPATH/src/github.com/dannyvankooten && \
+    ln -s /app $GOPATH/src/github.com/dannyvankooten/browserpass
+
+WORKDIR /app
```
maximbaz: could we name it `browserpass` instead of `app`? I might have hardcoded the folder name in the makefile :smile:

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.8-alpine3.6
+
+ENV CGO_ENABLED=0 \
+    PATH="$PATH:/app/node_modules/.bin/"
+
+RUN apk add --no-cache bash make zip chromium yarn && \
+    yarn install && \
```
maximbaz: You need to delete this line. As I mentioned earlier, we have a `make deps` now, and user will use it when actually running a container. Fetching dependencies (like `yarn install` or `dep ensure`) needs to be done at runtime, not at image build time, because they change from time to time and we don't want to ask people to recompile the image.

jayme-github:
Dropped the chrome stuff for now as i'ts probably not worth it. Other targets work fine for me now.

maximbaz:
I tested this on top of #149, everything seems to work fine for me too - thanks!

I'll get #149 merged first, then this one, will take care of conflicts in CONTRIBUTING.md myself (if any).

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.8-alpine3.6
```
jannickfahlbusch: I think it would be better if you would use `golang:alpine` which points to the latest Go version.

jayme-github: I like pinning of versions because you have a defined state that way. Without version tags people end up with different build environments depending on when they did a docker build.
I would vote for keeping it the way it is.

jannickfahlbusch: That is a point which I didn't thought of. With that in mind, I'd also go your way.

maximbaz:
Merged in https://github.com/dannyvankooten/browserpass/commit/f5eaf01ffafa3224d0dbd9c8691e0df37e58c52c, thanks!

PR #145, opened 2017-09-19
Add issue template

chrboe:
Fixes #145

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,19 @@
+### General information
+<!-- Please answer all the questions below before submitting the issue. -->
+
+* Operating system + version:
+* Browser + version:
+* If installed an official release, put a release version: 
+* If built from sources, put a commit id (`$ git describe --always`): 
+* Browserpass extension version as reported by your browser:
+
+---
+
+### Exact steps to reproduce the problem
+1.
+2.
+3.
+
+### What should happen?
+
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+### General information
+*(Please answer all the questions below and delete this line before submitting the issue.)*
```
maximbaz: instead of asking to delete this line, wrap it in html comment, like `<!-- Please answer ..... -->`

chrboe: Woah, I didn't know that was possible 😮

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+* Operating system + version:
+* Browser + version:
+* Built from source?
+* Browserpass version or commit if built from source:
```
maximbaz: I guess most people will use a release, so I propose to replace the last two questions with the following:

```
* If installed an official release, put a release version: 
* If built from sources, put a commit id (`$ git describe --always`): 
* Browserpass extension version as reported by your browser:
```

The last question I put is because many people still use the old 1.0.5, what is available in Webstore.

chrboe: Seems sensible to me. I knew the wording was a bit awkward, but in lack of a better solution I decided to roll with it anyways.

Issue #136, opened 2017-09-16
RFC: Guess config directory in install.sh?


chrboe:
I was wondering if we could just "guess" the config directory for each browser (as in, `[[ -d ~./config/google-chrome ]] # put extension there`) 

The upside would be that
1. We could get rid of the browser selection prompt (or replace it with `Chrome detected, install for Chrome? [y/n]` type prompts)
2. We could properly detect development/canary/testing versions (#110) by just adding the config path to an array
3. People who use multiple browsers (for whatever reason) don't have to execute the script twice

The only downside I can see is that people might only want to install the extension for one particular browser, but that could be fixed by the prompts suggested in 1.

maximbaz:
I like the idea, ideally have a map defined somewhere on top with browser names and their locations (so that it's easy to add new entries), and then a generic loop for all entries doing the work. Prompt is a good idea, let's do it.

qbit:
I think it would make sense to do this.

chrboe:
Little update on this: I've implemented a rough draft of how this could look, but I'm unable to test it right now. Once I get my hands on a Linux machine again I'll try it out and submit a PR.

maximbaz:
This install script has been mentioned in a few other tickets recently, and remembering about this ticket and the fact that we had some issues with discovering browsers based on existence of configuration folders (details in #147), I'm wondering if this is really worth doing after all.

Point number 3 (having to restart script for multiple browsers) is a valid flaw, but maybe we should just allow typing multiple digits at the same time? 

```
Select your browser:
====================
1) Chrome
2) Chromium
3) Firefox
4) Vivaldi
1-4: 1 3 <Enter>
```

If we can refactor the code to simplify adding more browsers, that of course is nice too.

maximbaz:
I think I'm going to close this and the PR due to the reasons above, let me know if you want to follow-up on any of the ideas above.

PR #147, opened 2017-09-19
Guess config directories in install.sh

chrboe:
Tested only on Linux with Chromium due to time constraints (and because I don't have a Mac at hand 😜). Feel free to test on other platforms.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -123,12 +123,12 @@ for browser in "${BROWSER_PATHS[@]}" ; do
     browser_name=${browser%%:*}
     browser_path=${browser#*:}
 
-    if [ -d $(dirname "$browser_path") ]; then
+    if [ -d "$(dirname "$browser_path")" ]; then
         no_browser_found=false
         read -r -p "Detected $browser_name. Install for this browser? [Y/n] " response
         case "$response" in
             [yY][eE][sS]|[yY])
-                install_host_config $browser_name $browser_path
+                install_host_config "$browser_name" "$browser_path"
                 ;;
             *)
                 rejected_installs+=("$(dirname "$browser_path")")
--- a/install.sh
+++ b/install.sh
@@ -6,141 +6,142 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
 APP_NAME="com.dannyvankooten.browserpass"
 HOST_FILE="$DIR/browserpass"
 
+install_host_config () {
+    browser_name=$1
+    target_dir=$2
+
+    echo "Installing $browser_name host config"
+
+    # Create config dir if not existing
+    mkdir -p "$target_dir"
+
+    # Escape host file
+    ESCAPED_HOST_FILE=${HOST_FILE////\\/}
+
+
+    # Copy manifest host config file
+    if [ "$browser_name" == "Chrome" ] || \
+        [ "$browser_name" == "Chromium" ] || \
+        [ "$browser_name" == "Vivaldi" ]; then
+        if [ ! -f "$DIR/chrome-host.json" ]; then
+            echo "ERROR: '$DIR/chrome-host.json' is missing."
+            echo "If you are running './install.sh' from a release archive, please file a bug."
+            echo "If you are running './install.sh' from the source code, make sure to follow CONTRIBUTING.md on how to build first."
+            exit 1
+        fi
+        cp "$DIR/chrome-host.json" "$target_dir/$APP_NAME.json"
+    else
+        if [ ! -f "$DIR/firefox-host.json" ]; then
+            echo "ERROR: '$DIR/firefox-host.json' is missing."
+            echo "If you are running './install.sh' from a release archive, please file a bug."
+            echo "If you are running './install.sh' from the source code, make sure to follow CONTRIBUTING.md on how to build first."
+            exit 1
+        fi
+        cp "$DIR/firefox-host.json" "$target_dir/$APP_NAME.json"
+    fi
+
+    # Replace path to host
+    sed -i -e "s/%%replace%%/$ESCAPED_HOST_FILE/" "$target_dir/$APP_NAME.json"
+
+    # Set permissions for the manifest so that all users can read it.
+    chmod o+r "$target_dir/$APP_NAME.json"
+
+    echo "Native messaging host for $browser_name has been installed to $target_dir."
+}
+
 # Find target dirs for various browsers & OS'es
 # https://developer.chrome.com/extensions/nativeMessaging#native-messaging-host-location
 # https://wiki.mozilla.org/WebExtensions/Native_Messaging
 OPERATING_SYSTEM=$(uname -s)
 
 case $OPERATING_SYSTEM in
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -6,141 +6,142 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
...
-  fi
-  cp "$DIR/chrome-host.json" "$TARGET_DIR/$APP_NAME.json"
-else
-  if [ ! -f "$DIR/firefox-host.json" ]; then
-    echo "ERROR: '$DIR/firefox-host.json' is missing."
-    echo "If you are running './install.sh' from a release archive, please file a bug."
-    echo "If you are running './install.sh' from the source code, make sure to follow CONTRIBUTING.md on how to build first."
+if [ "$no_browser_found" = true ]; then
+    echo "No compatible browsers found."
+    echo "If  you do  actually have a  browser installed  which you  think should  be supported,"
```
maximbaz: Is this a simulation of Word's "text align - justify" here? 😄

install.sh
```diff
@@ -6,141 +6,142 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
...
-  echo "$OPERATING_SYSTEM is not supported"
-  exit 1
-  ;;
+    Linux)
+        HOST_FILE="$DIR/browserpass-linux64"
+        if [ "$(whoami)" == "root" ]; then
+            BROWSER_PATHS=( "Chrome:/etc/opt/chrome/native-messaging-hosts"
+            "Chromium:/etc/chromium/native-messaging-hosts"
+            "Firefox:/usr/lib/mozilla/native-messaging-hosts"
+            "Vivaldi:/etc/chromium/native-messaging-hosts" )
```
maximbaz: Chromium and Vivaldi have the same folder, so Vivaldi is being falsely discovered as available. Do you think it is a copy-paste issue that nobody has discovered so far? 😄

maximbaz:
So although I only have Chromium, your script discovered Chrome, Chromium and Firefox as available 🙂 

There is a good reason for that, I had them installed at some point, but although I deleted them, the configuration folders remained.

I'm concerned that we will be getting many new issues like "Hey, your install.sh says I have Firefox installed although I do not". As an idea, if user answers "no", maybe we can print a message, something like "If you feel this suggestion was a mistake and you don't have this browser installed, delete this folder: \<full-path\>"?

maximbaz:
Also, how hard would it be to do the same for Powershell script, so we can fully close the #136? /cc @qbit

qbit:
mm I guess it depends on the other browsers and what location they use to store the registry stuff. Shouldn't be much harder than determining the locations.

maximbaz:
Is it possible to use the same approach as here, i.e. since all Powershell does is writing into registry as `hk{0}:\Software\BROWSER_NAME\NativeMessagingHosts`, just check if registry key `hk{0}:\Software\BROWSER_NAME` exists, and if yes, prompt "Detected BROWSER_NAME, install for this browser? [Y/n]"

install.sh
chrboe: Yep 😄 It really bothered me how the line ends were *almost* aligned, but not quite. Idk, it was a bit of an experiment. If you think it looks ugly I can still revert it 🤷‍♂️

install.sh
chrboe: Wow, I hadn't actually noticed that. It's literally been there [forever](https://github.com/dannyvankooten/browserpass/commit/9a313be3ad53ecdfb4bc53dd1c2543af5570038f#diff-3fbb47e318cd8802bd325e7da9aaabe8R31) though 😄 

I'll follow up with a fix sometime today.

chrboe:
@maximbaz 

> I had them installed at some point, but although I deleted them, the configuration folders remained.

That's weird, shouldn't the package manager take care of deleting the config directory? 🤔
I guess the solution you propose would work just fine, but it kind of seems like a botch to me (from a user perspective). If we had an FAQ section I guess we could put an explanation there and print a link to it in `install.sh`.

Also, since I'm an Arch user myself, it kind of worries me how pacman just leaves stale config directories behind 😄 Do you happen to know whether or not there's a good reason for that?

Anyways, in the meantime I'll implement your proposed solution.

install.sh
maximbaz: It looks a little bit weird to me, let's revert 🙂

maximbaz:
Yeah it makes sense indeed. 

I'm fine with both approaches, either print such warning directly from the script, or add this explanation to a README.md and refer to that section from the script. I'd choose depending on the size of the message.

maximbaz:
Cool cool :+1: Are you planning on doing anything else here, or it is ready to be merged?

I'm planning on doing some final tests and merging all open PRs tomorrow in the afternoon.

chrboe:
From my point of view I'm done here. The only thing I'm not 100% certain about is the global Vivaldi path. I looked around on the internet but couldn't find a definite source on whether or not `/etc/vivaldi` is actually correct.

It would be nice if someone who has Vivaldi installed could test installing with root (or even just check if the path is correct)

maximbaz:
So I installed Vivaldi, and my experiments show that it doesn't read from `/etc` at all, neither from `/etc/vivaldi`, nor from `/etc/Vivaldi`, nor from `/etc/chromium`... It only reads from `$HOME/.config/vivaldi`.

maximbaz:
Another test that I made:

- delete chromium entirely
- delete `/etc/chromium`
- install chromium
- check if `/etc/chromium` exists - it does not. So `sudo ./install.sh` doesn't detect that I have chromium installed.

So I don't like this approach of checking the existence of the config folders for detecting if a browser is available... Maybe we can do something else instead? Just a wild idea, maybe detect the existence of binaries in `$PATH`?

chrboe:
Dang, that sucks... I guess *discovering* the browser would work using the `$PATH` method (or `whereis`?), but what would we do after that? Just create the config directory on behalf of the browser if it doesn't exist?

maximbaz:
That's what we do today, actually :)

https://github.com/dannyvankooten/browserpass/blob/d5fc4fddf1d7a21455ab342e80d6472fe623b287/install.sh#L113-L114

Like I'm pretty sure I haven't had `/etc/chromium` at all until I started using browserpass.

I'm more concerted of how we implement the discovery mechanism, have you seen how many binaries the makefile tries just for Chrome and Chromium? :) I hope Firefox and others have less potential names :)

https://github.com/dannyvankooten/browserpass/blob/d5fc4fddf1d7a21455ab342e80d6472fe623b287/makefile#L2

chrboe:
@maximbaz just wanted to give an update on this PR real quick. I'm pretty busy right now with university starting again next week, so I haven't had much time to work on browserpass. I will definitely come back to this patch very soon™️ though!

maximbaz:
Sounds good :+1:

maximbaz:
Closing due to https://github.com/dannyvankooten/browserpass/issues/136#issuecomment-353999726

PR #148, opened 2017-09-19
Build release src tarball (and remove vendor)

qbit:
- Update build process docs
- add two new targets in the makefile: deps, tarball
- always run "release" when running "sign"
- ignore the vendor directory


maximbaz:
Thanks a lot! I merged in https://github.com/dannyvankooten/browserpass/commit/ddefdd68e98c94e7706ccce40f30ac58849412b5, also made a few adjustments to the `tarball` goal, namely removing the `.git` folder (I don't think package managers need it, right?) as well as `*.pem` files 😄

PR #149, opened 2017-09-19
Simplify usage of make

maximbaz:
* All variants of Chrome binaries are already defined in makefile, no need to change it.
* Developers should be concerned only of two make goals, `make js` and `make browserpass`
    * /cc @jayme-github - this is what mostly needed from Docker to support
    * `make release` doesn't make sense to be used for regular people, as without my `*.pem` file extension will get a different ID and host app will not work with it.
* Dependencies will always be fetched automatically.
* Package maintainers should not be using `make` anymore as browserpass-src.tar.gz (made with `make tarball`) contains everything.

@qbit, @chrboe - let me know if this change makes sense to you, as I'm still not very comfortable with makefiles.

### DIFF
--- a/makefile
+++ b/makefile
@@ -3,11 +3,10 @@ CHROME := $(shell which google-chrome 2>/dev/null || which google-chrome-stable 
 PEM := $(shell find . -name "chrome-browserpass.pem")
 JS_OUTPUT := chrome/script.js chrome/inject.js
 
-.PHONY: empty
-empty:
+all: js browserpass
 
-.PHONY: chrome
-chrome:
+.PHONY: crx
+crx:
 ifeq ($(PEM),./chrome-browserpass.pem)
 	"$(CHROME)" --disable-gpu --pack-extension=./chrome --pack-extension-key=$(PEM)
 else
@@ -16,14 +15,11 @@ else
 endif
 	mv chrome.crx chrome-browserpass.crx
 
-.PHONY: firefox
-firefox:
-	cp chrome/{*.html,*.css,*.js,*.png,*.svg} firefox/
-
 .PHONY: js
 js: deps $(JS_OUTPUT)
 	cp chrome/host.json chrome-host.json
 	cp firefox/host.json firefox-host.json
+	cp chrome/{*.html,*.css,*.js,*.png,*.svg} firefox/
 
 chrome/script.js: chrome/script.browserify.js
 	browserify chrome/script.browserify.js -o chrome/script.js
@@ -50,6 +46,7 @@ browserpass-freebsd64: deps cmd/browserpass/ pass/ browserpass.go
 	env GOOS=freebsd GOARCH=amd64 go build -o $@ ./cmd/browserpass
 
 clean:
+	rm -f browserpass
 	rm -f browserpass-*
 	rm -rf release
 	git clean -fdx chrome/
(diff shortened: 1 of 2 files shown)


qbit:
Looks good to me. I'd like to see `all` get defined so we could just run `make` (`gmake` in my case) and have it do TheRightThing™. It would also be super cool if we could only run `dep` or `yarn` if the `node_modules`/`vendor` directories were missing.. but I am not sure how that could be done.

makefile
```diff
@@ -21,35 +21,32 @@ firefox:
...
 
-browserpass-openbsd64: cmd/browserpass/ pass/ browserpass.go
+browserpass-openbsd64: deps cmd/browserpass/ pass/ browserpass.go
 	env GOOS=openbsd GOARCH=amd64 go build -o $@ ./cmd/browserpass
 
-browserpass-freebsd64: cmd/browserpass/ pass/ browserpass.go
+browserpass-freebsd64: deps cmd/browserpass/ pass/ browserpass.go
 	env GOOS=freebsd GOARCH=amd64 go build -o $@ ./cmd/browserpass
 
 clean:
```
jayme-github: We should add `rm -f browserpass` to the clean target as well.

chrboe:
I agree with @qbit, an `all` target would be nice.

Regarding
> It would also be super cool if we could only run dep or yarn if the node_modules/vendor directories were missing

yarn actually checks this by itself, so it effectively no-ops when there's nothing to do. A short test confirms this behaviour:

```
$ yarn
yarn install v1.0.2
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
Done in 4.85s.

$ yarn
yarn install v1.0.2
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.42s.
```

I'm not sure about `dep`, but I'd expect the same behaviour there.

Of course, if we absolutely wanted to do the check (saving 0.42s), we could do `[[ ! -d "node_modules" ]] && yarn`, but that wouldn't actually update the dependencies after a `git pull`, for instance.

jayme-github:
`dep ensure` is quite fast when dependencies have already been pulled. Checking for existence of `node_modules` and `vendor` directories is indeed a bad idea.

chrboe:
Regarding the `all` target: it might actually be as simple as doing something like this

```
diff --git a/makefile b/makefile
index ee347a0..53e74b4 100644
--- a/makefile
+++ b/makefile
@@ -3,8 +3,7 @@ CHROME := $(shell which google-chrome 2>/dev/null || which google-chrome-stable
 PEM := $(shell find . -name "chrome-browserpass.pem")
 JS_OUTPUT := chrome/script.js chrome/inject.js

-.PHONY: empty
-empty:
+all: browserpass js

 .PHONY: chrome
 chrome:
```

makefile
maximbaz: Nice catch, thanks

maximbaz:
Thanks to all of you!

The new commit adds the following:
- Added `all` goal so that running `make` just works
- Merged the `firefox` goal into `js`, so that `js` prepares both Chrome and Firefox unpacked extensions at once (also extended documentation in CONTRIBUTING.md about this)
- Renamed `chrome` to `crx` make goal

Running `dep` / `yarn` only if there is no `vendor` / `node_modules` is a bad idea because from time to time I will be updating dependencies, and when I do so and commit `Gopkg.lock` / `yarn.lock`, you want to also update the dependencies on your side. Running `dep` / `yarn` does exactly that, when there are no changes to the lock files, they do nothing, when there are changes to the lock files, they pull new dependencies.

Let me know if you think this change is ready to be merged.

makefile
```diff
@@ -16,14 +15,11 @@ else
 endif
 	mv chrome.crx chrome-browserpass.crx
 
-.PHONY: firefox
-firefox:
-	cp chrome/{*.html,*.css,*.js,*.png,*.svg} firefox/
-
 .PHONY: js
 js: deps $(JS_OUTPUT)
```
jayme-github: If `$(JS_OUTPUT)` files have been build but `{chrome,firefox}-host.json` and/or `firefox/{*.html,*.css,*.js,*.png,*.svg}` have not, this target will exit with `Nothing to do for "js"`.

jayme-github: Or, im wrong because js is a phony target. In that case: sorry for the noise

maximbaz: Tested this, to be sure:

```
$ make js
$ rm chrome-host.json
$ make js
```

The `chrome-host.json` reappeared, so seems to be fine 👍

Issue #40, opened 2017-02-10
Enter key to autocomplete


emersion:
_Just do it_ when the enter key is pressed.

dannyvankooten:
Hey @emersion,

You mean when there's only result? Because in the current set-up, you should be able to cycle through the search results using TAB and then hit ENTER to submit login form with the selected login details.

emersion:
It would be cool to submit the first result when the enter key is pressed.

Another idea of useful keybinding would be to be able to cycle through results using arrow up/down.

dannyvankooten:
The issue I have with immediately submitting the first result on ENTER is that is has to be super clear that the search field lost focus, otherwise people may end up sending login credentials to the wrong site when they meant to simply issue another search. There shouldn't be too much magic there, imo.

I can definitely see up/down arrows (in addition to TAB) being helpful though. :1st_place_medal:

FabianHenneke:
A safer way to implement this could be to only fill on ENTER if the current domain is still the one extracted from the tab's URL.

maximbaz:
I think there is a simple way of resolving this issue:

When a user hits shortcut to open the popup for the first time, if there is any match at all, the focus should be on the first match, and not on the search box. If there are no matches, the cursor stays on the search box for the user to type some request.

madduck:
Even if there are multiple matches, it should be easy to just go for the first, even if the search box is highlighted. I think visually, this is best done by selecting the first entry (highlighting it), leaving the cursor in the search box, but letting up/down move the selection around, while Enter then simply picks the selected item.

This will be especially necessary in the context of #32 (i.e. fuzzy searching).

maximbaz:
This might be a bit confusing, because hitting <kbd>Enter</kbd> while cursor is in the search box can mean two things: "do the search" and "submit the first entry".

For clarity from the user perspective I'd prefer keeping focus on the thing that <kbd>Enter</kbd> is going to affect - if cursor is on the search box, do the search, if cursor is on a focused entry, select it.

erayd:
How about the following?

 * If the search box is empty, enter submits the first entry.
 * If the search box contains text, then:
    + If the search has not yet been run, run the search
    + If the search has been run, and has not been edited since running, submit the first entry.

This seems like it would be a fairly intuitive workflow, and allow using enter for both search initiation and submitting results for autofill.

erayd:
Or an alternative approach could be:

[...]

PR #152, opened 2017-09-23
Keyboar selection in list using arrow keys

turboMaCk:
This is still work in progress (missing FF version).

I've really missed arrow keys selection so I've coded it myself. I'm opening this PR early in case you're interested to review and discuss it. If you're not interested in this feature, feel free to close this PR - I can always use my own version of extension anyway. In case you're interested I'm going to clean code a bit implement the same feature for firefox as well as include changes from code review.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -174,6 +174,6 @@ function switchFocus(firstSelector, nextNodeAttr) {
   if (newActive) {
     newActive.focus();
   } else {
-    document.getElementById('search-field').focus();
+    document.getElementById(inputId).focus();
   }
 }
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -5,7 +5,6 @@ var app = "com.dannyvankooten.browserpass";
 var activeTab;
 var searching = false;
 var logins = null;
-var highlightedItem = null;
 var domain, urlDuringSearch;
 
 m.mount(document.getElementById("mount"), { view: view });
@@ -19,11 +18,6 @@ chrome.tabs.query({ currentWindow: true, active: true }, function(tabs) {
 function view() {
   var results = "";
 
-  function style(faviconUrl, l, i) {
-    var style = `background-image: url('${faviconUrl}');`;
-    return i == highlightedItem ? style + ` background-color: yellow;` : style;
-  }
-
   if (searching) {
     results = m("div.loader");
   } else if (logins !== null) {
@@ -35,13 +29,13 @@ function view() {
         m.trust(`No passwords found for <strong>${domain}</strong>.`)
       );
     } else if (logins.length > 0) {
-      results = logins.map(function(l, i) {
+      results = logins.map(function(l) {
         var faviconUrl = getFaviconUrl(domain);
         return m(
           "button.login",
           {
             onclick: getLoginData.bind(l),
-            style: style(faviconUrl, l, i)
+            style: `background-image: url('${faviconUrl}')`
           },
           l
         );
@@ -51,20 +45,20 @@ function view() {
 
   return [
     // search form
-    m("div.search", [
+    m("div.search", { onkeydown: keyHandler }, [
       m(
         "form",
         {
-          onsubmit: submitSearchForm(results)
+          onsubmit: submitSearchForm
         },
         [
           m("input", {
             type: "text",
+            id: "search-field",
             name: "s",
             placeholder: "Search password..",
             autocomplete: "off",
-            autofocus: "on",
-            onkeydown: keyHandler.bind(results),
+            autofocus: "on"
           }),
           m("input", {
             type: "submit",
@@ -76,29 +70,19 @@ function view() {
     ]),
 
     // results
-    m("div.results", results)
+    m("div.results", { onkeydown: keyHandler }, results)
   ];
 }
 
-function submitSearchForm(resultsArray) {
-  return function (e) {
-    e.preventDefault();
+function submitSearchForm(e) {
+  e.preventDefault();
 
-    // don't search without input.
-    if (!this.s.value.length) {
-
-      // open highlighted item
-      if (highlightedItem !== null) {
-        var item = resultsArray[highlightedItem];
-        if (item) {
-          return getLoginData.call(item.children[0]);
-        }
-      }
-      return;
-    }
-
-    searchPassword(this.s.value);
+  // don't search without input.
+  if (!this.s.value.length) {
+    return;
   }
+
+  searchPassword(this.s.value);
 }
 
 function init(tab) {
@@ -156,7 +140,6 @@ function getFaviconUrl(domain) {
 function getLoginData() {
   searching = true;
   logins = null;
-  highlightedItem = null;
   m.redraw();
 
   chrome.runtime.sendMessage(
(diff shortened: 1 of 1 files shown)


maximbaz:
Thanks @turboMaCk, I'm definitely interested in getting this merged 🙂 This feature was also requested a while ago in a [comment](https://github.com/dannyvankooten/browserpass/issues/40#issuecomment-292882196) somewhere else.

I'll be reviewing the code a bit later today or tomorrow.

maximbaz:
I've tried it out, my immediate thoughts: I'm concerned that you are basically re-implementing the selection feature, you are writing lots of code to do this, you are coloring the selection, keeping track of highlighted item, etc... First of all, all this is already implemented by the browser itself (what you see when pressing `Tab` and `Shift+Tab`), second - it is conflicting with pressing `Tab` (try pressing both arrows and `Tab`, you will see two selections).

I have two ideas, not sure if any of them is possible, but could potentially simplify this PR a lot.

1. Add `keydown` listener, and when `Up` or `Down` is pressed, just initiate a `Tab` or `Shift+Tab` keypress, and let the browser do the rest of the work.
2. Similar to above, but instead of initiating `Tab` keypress, manipulate `tabIndex` property of the buttons. A bit trickier, but still simpler, just find the currently focused element and shift the index to the next one.

Both ideas will allow you to reuse most of what is already implemented by browsers, and arrows will not conflict with `Tab` / `Shift+Tab` key presses.

What do you think about these?

maximbaz:
Had a chance to investigate a bit, here are the findings:

1. Browsers do not allow triggering Tab keypress, so this is not an option for us.
2. This will work, and it is actually relatively simple. `document.activeElement` returns you an element which currently has focus. You have several possible outcomes here:
    - The focus is currently on the search input box - this is the starting point. In such case, find the first button and call `.focus()` on it.
    - A button has focus - use the `document.activeElement.nextElementSibling` to find the next button:
        - If it is not null, call `.focus()` on it
        - If it is null, it means the last button has focus now - move the focus back to the input box.
    - Nothing has focus - don't think this is possible, but you can handle this as well, just in case, and move the focus to the input box.

turboMaCk:
Hello, @maximbaz :wave: 

Thanks for super fast response. I was away until now so sorry for not keeping up with you in this.

Frankly I even didn't notice that I can use native focus with this extension since I dind't know items are buttons but it makes quite some sense. There are some advantages for using custom highlighting over native focus. For instance I like to be able to type and using sellection at the same time but probably it's not as usefull in this case sice search is performed on form submit not input (which makes sense from security perspective).

The fact that this doesn't play well with native focus is game changer indeed though and I agree it we need to fix that. In this implementation key handler is on input so on `Tab` you even loose that and there are other issues like ones you have already described.

I think it's pretty clear that you would prefer arrows to behave as much like native focus as possible so I'll reimplement it that way. One thing I'm not so much sure about is how will usage of `DOM` apis play with virtual dom of mithril. Probably the best way to find out is to try most simple solution first and see how it goes.

I'll keep you up to date. Let me know if you have so additional points.

maximbaz:
Yeah I'd prefer to use the native focus as much as possible. Let me know if you get into issues and I'll try to help as much as I can, but I tested the things above (like `document.activeElement` and `.focus()`) in console and they did work fine.

By the way, I was toying recently with the idea to perform search on keydown instead of form submit, and even if we ignore security perspective for a minute, the current UI is simply not fit for this, everything is rather ugly, the box is jumping and changing width and height every time, and flashing between doing the searches... in the end, it's no good, I think there are too many reasons for now to stay with the "form submit" event 🙂

chrome/script.browserify.js
```diff
@@ -54,6 +54,7 @@ function view() {
         [
           m("input", {
             type: "text",
+            id: "search-field",
```
turboMaCk: I'm using id for selecting and detecting focus on `input`. It seemed to be better than reading other attributes (like tagName) to me.

chrome/script.browserify.js
```diff
@@ -69,7 +70,7 @@ function view() {
     ]),
 
     // results
-    m("div.results", results)
+    m("div.results", { onkeydown: keyHandler }, results)
```
turboMaCk: As you can see handler is assigned to a container of results and container of input. I did it because I didn't want to change structure without asking you first. Alternatively, we can add a new extra container to render or use native API to register the handler.

chrome/script.browserify.js
```diff
@@ -149,3 +150,30 @@ function getLoginData() {
...
+  case 'ArrowDown':
+    switchFocus('button.login:first-child', 'nextElementSibling');
+    break;
+  }
+}
+
+function switchFocus(firstSelector, nextNodeAttr) {
+  var inputId = 'search-field';
+  var newActive = document.activeElement.id === inputId ?
+      document.querySelector(firstSelector) :
```
turboMaCk: this is using querySelector. Compatibility should be probably fine due to browser selection. https://www.w3schools.com/jsref/met_document_queryselector.asp

turboMaCk:
I'll wait for your review before porting/integrating this to FF version.

chrome/script.browserify.js
maximbaz: I don't mind wrapping both containers in a some kind of parent container, and assigning the onkeydown to it. But if for whatever reason it gets complicated, the current solution is also fine.

maximbaz:
I like it much better now @turboMaCk! And by the way, it works very well on both my Chromium and Firefox 😉 Just to make sure, you know that we have the same code for both browsers, right? After you run `make js`, the `firefox` folder is ready to be imported [as temporary extension](https://github.com/dannyvankooten/browserpass/blob/master/CONTRIBUTING.md#to-load-an-unpacked-extension). Do test on your FF as well though, just to make sure it works for you too 🙂

Regarding to your suggestion to wrap everything in a parent container, seems like a valid idea to me. Try it out, and let me know when you are done implementing and testing, so that I can do one final test and merge everything 🙂

turboMaCk:
Gotcha. There is one tiny issue in FireFox though. Native focus circles between `input > [ list of items one by one ] > **whole form**. This causes 2 things:

- Arrows behave differently (never focus the whole form and goes straight to input).
- When the whole form is focused, Arrows are not working.

2nd issue should be possible to fix by reversing logic in [this check](https://github.com/dannyvankooten/browserpass/pull/152/files#diff-94860ae26dd8634c502436020a3e0453R170) to not `button.item` I think.

The first issue would require overloading `Tab` behavior which seems hacky.

chrome/script.browserify.js
```diff
@@ -43,34 +43,35 @@ function view() {
...
-            style: "display: none;"
-          })
-        ]
-      )
-    ]),
-
-    // results
-    m("div.results", results)
-  ];
+  return m('div.container', { onkeydown: keyHandler }, [
```
turboMaCk: stupid class name but is fine for the view as small as this.

maximbaz:
1. Let's not overload `Tab` behavior, what you implemented for arrows is good and intuitive.
2. The chance of people using both `Tab` and arrows is small, but if you have an idea on how to fix arrows while the entire form is focused - let's try it.

turboMaCk:
It seems that extra thing firefox focus is not `form` but some node above new `div.container`. This means that even event listener is not firing when it's focused. Solution to that would be probably to register a listener on the whole document but that seems to me like anti-pattern in Mithril. Let me know what you think but to me, this seems to be good enough. We can always add any hack later in case users will find it confusing.

maximbaz:
Then let's not introduce any such hacks for now. I'll test it later today, and merge if everything is okay. Thanks for this PR!

maximbaz:
Merged in https://github.com/dannyvankooten/browserpass/commit/c3a6279f145e4fa03cbc6db426521279bdf6dfc.

What I did extra is I removed a parent ".container", because it looked like that and I don't see a need for the outermost container:

image

Thanks once again!

-----

My testing on Firefox revealed the following troubling issue: Firefox does not respect the [`autofocus` property](https://github.com/dannyvankooten/browserpass/blob/c3a6279f145e4fa03cbc6db426521279bdf6dfc2/chrome/script.browserify.js#L61), meaning that when users use hotkey to open the popup, they still need to press `Tab, Tab` to even get the focus to the search box. This is not related to your change, I'm just surprised that nobody reported this issue so far. Once the focus is on the search box, arrows work as expected.

turboMaCk:
@maximbaz I've spotted that issue as well myself just didn't want to mix it with this one. Great you've made an issue - I wanted to create one myself once this is closed. I'll subscribe to that issue and in case I'll have some time before anybody else will pick it up I'll do so myself.

Thanks for all your help with this PR - very appreciated.

PR #153, opened 2017-09-24
Add Security section to README

### LINKED ISSUE #146: Publish and commit to security goals
The reason I would personally use an extension instead of `-c` is that it can extract the hostname securely, blocking phishing attempts. The reason I would personally use this instead of, say, LastPass is that there would be no way for the page to trigger autocomplete (and exploit potential host-matching vulnerabilities) without user input and confirmation.

These are however only my speculations on the security goals of the extension. I could audit it myself, but I would have to worry about functionality changing later as there is no commitment. Also, that would make clear what is and what isn't a security vulnerability.

Essentially, *what's your threat model*? (I know it's cliche.)

(Thanks for the work! ✨)

FiloSottile:
These principles make phishing prevention a stated goal and mitigate potential vulnerabilities as seen in other password managers:

https://bugs.chromium.org/p/project-zero/issues/list?can=1&q=label:Product-LastPass

Closes #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,15 @@ Click the lock icon or use **Alt + Shift + L** to fill & submit your login info 
 
 _Note: This does not work in Firefox before version 52._
 
+## Security
+
+Browserpass aims to protect your passwords and computer from malicious or fraudulent websites.
+
+* To protect against phishing, only passwords matching the origin hostname are suggested or selected without an explicit search term.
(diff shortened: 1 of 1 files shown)


maximbaz:
LGTM. @qbit want to have a look or suggest some additions before I merge?

qbit:
Looks good to me!

PR #154, opened 2017-09-24
Display OTP label if we have one.

qbit:


chrome/inject.browserify.js
```diff
@@ -77,7 +77,7 @@
   update(field(USERNAME_FIELDS), login.u);
 
   if (login.digits) {
-    alert(login.digits);
+    alert(login.label != "" ? login.label + ": " + login.digits : login.digits);
```
maximbaz: How about this?

`alert((login.label || "OTP") + ": " + login.digits);`

qbit: Makes sense to me - updated.

maximbaz:
Thanks as usual 🙂

Issue #155, opened 2017-09-25
The search box is not autofocused in FF when popup is opened


maximbaz:
### General information

* Operating system + version: ArchLinux
* Browser + version: FF 55.0.3
* If installed an official release, put a release version: 
* If built from sources, put a commit id (`$ git describe --always`): https://github.com/dannyvankooten/browserpass/commit/c3a6279f145e4fa03cbc6db426521279bdf6dfc2
* Browserpass extension version as reported by your browser: 1.0.15

---

### Exact steps to reproduce the problem
1. Install the extension on Firefox
2. Hit the hotkey <kbd>Alt+Shift+L</kbd>

### What should happen?

The search input box should autofocus, ready to receive search term, just like on Chrome.

### What happened instead?

Need to press <kbd>Tab</kbd>, <kbd>Tab</kbd> to get to the search box.

-----

May be somewhat related to #40.

turboMaCk:
I think we can use [this](https://mithril.js.org/lifecycle-methods.html#oncreate) lifecycle hook on the very top component to make a check if the input is selected. In case it's not we can use standard `DOM` API to do select via script. What do you think?

turboMaCk:
Also, I agree that this https://github.com/dannyvankooten/browserpass/issues/40#issuecomment-301186768 feature would be nice. Anyway, I would suggest to do it one by one. First, start with solving autofocus issue and then adding detection about what should be focused once popup is opened.

maximbaz:
I'm fine with doing it one by one :+1:

turboMaCk:
I'll start working on it after dinner then. I think it won't take much time.

maximbaz:
Awesome job with this bug :+1: 
Wanna tackle https://github.com/dannyvankooten/browserpass/issues/40#issuecomment-301186768 next? There's no rush, whenever you have time 🙂

phryneas:
Hey there, I was just experiencing the same problem and the mailing list led me to this open Issue in the Firefox Bug Tracker - you might want to monitor that too :)
https://bugzilla.mozilla.org/show_bug.cgi?id=1324255

PR #158, opened 2017-09-27
WIP: Bug/ff input auto select

turboMaCk:
issue: https://github.com/dannyvankooten/browserpass/issues/155

This is an initial spike of a bugfix for FireFox which is not autofocusing search input when the popup is opened.

I've had to upgrade mithril since old version [doesn't provide](https://github.com/MithrilJS/mithril.js/tree/v0_2_x) lifecycle hooks we need. It seems there is no change in API that this extension is using.

Secondly, it turned out that FireFox is ignoring `focus` calls even though the element is already in DOM and can be queried via DOM API. This is the reason for `setTimeout`.

Finally, it seems I have a different version in my fork than is in upstream (even though I tried to pull) - I'll resolve it.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -167,7 +167,7 @@ function keyHandler(e) {
 
 function switchFocus(firstSelector, nextNodeAttr) {
   var searchField = document.getElementById('search-field');
-  var newActive = document.activeElement !== searchField
+  var newActive = document.activeElement === searchField
       ? document.querySelector(firstSelector)
       : document.activeElement[nextNodeAttr];
 
@@ -185,6 +185,7 @@ function oncreate() {
 
   // FireFox probably prevents `focus()` calls for some time
   // after extension is rendered.
+  searchField.focus();
   window.setTimeout(function() {
     if (searchField) {
       searchField.focus();
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -7,7 +7,7 @@ var searching = false;
 var logins = null;
 var domain, urlDuringSearch;
 
-m.mount(document.getElementById("mount"), { view: view });
+m.mount(document.getElementById("mount"), { view: view, oncreate: oncreate });
 
 chrome.browserAction.setIcon({ path: "icon-lock.svg" });
 chrome.tabs.onActivated.addListener(init);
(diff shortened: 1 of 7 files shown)


chrome/script.browserify.js
```diff
@@ -166,15 +166,28 @@ function keyHandler(e) {
...
+  var searchField = document.getElementById('search-field');
+  var newActive = document.activeElement !== searchField
       ? document.querySelector(firstSelector)
       : document.activeElement[nextNodeAttr];
 
   if (newActive) {
     newActive.focus();
   } else {
-    document.getElementById(inputId).focus();
+    searchField.focus();
```
turboMaCk: This is just a tiny change I forgot to push to the previous PR it seems. I can remove it from this one if you want.

chrome/script.browserify.js
```diff
@@ -166,15 +166,28 @@ function keyHandler(e) {
...
+function oncreate() {
+  var searchField = document.getElementById('search-field');
+
+  // FireFox probably prevents `focus()` calls for some time
+  // after extension is rendered.
+  window.setTimeout(function() {
+    if (searchField) {
+      searchField.focus();
+    }
+  }, 100);
```
turboMaCk: there is that ugly hack with timeout:(

chrome/script.browserify.js
maximbaz: No need, keep it here :+1:

chrome/script.browserify.js
maximbaz: Try to wrap it in `requestAnimationFrame` instead of the `setTimeout`, I'm curious if that's what FF wants.

maximbaz: I'd also move the call to `getElementById` inside the `requestAnimationFrame`, just to make sure we aren't searching too early :)

maximbaz:
I updated the all the dependencies on master, could you please rebase? 🙂

chrome/script.browserify.js
turboMaCk: I've just found some time to try it and it doesn't work which makes sense to me. if `window.setTimeout(fc, 0)` doesn't work I wouldn't expect `requestAnimationFrame` to work either. With `0` timeout browser evaluate function right after call-stack is empty. With the request, animation frame code is evaluated even before redraw.

I also tried this sort of "recursive" approach:

```js
function oncreate() {
  var searchField = document.getElementById('search-field');
  searchField.focus();

  if (document.activeElement !== searchField) {
    window.requestAnimationFrame(oncreate);
  }
}
```
which doesn't work either.

One thing I'm still not sure about is what makes timeout `100` works. Is it guaranteed to always work (since there is some timeout before `focus()` calls are accepted) or is some race in async code? I don't really know and frankly, I don't know how to test it either. So I did it just by trying. and it seems that **with 40ms timeout it sometimes works and sometimes doesn't**. Which seems to be answer - timeout might be simply buggy.

Next thing I did was to remove whole `setTimeout` hack together with `autofocus: "on"` and test that in chrome. And it worked correctly.

To me, it seems that same thing that prevents us to simply use `autofocus` in FF is the thing that blocks `focus()` call. Maybe we can try different DOM events but I'm skeptical.

Another solution that came to my mind was to rewrite that function so it will perform check every 10ms and if it's not successful it will set itself to run in next 10ms until it is. Anyway, I did quick and it seems that `document.activeElement` returns input even though it's not really selected so there is no (easy) way to perform that test.

maximbaz: Hey, thanks for doing the experiments! Then let's keep it simple and not over-engineer stuff, it's a bit hacky - yes, but it works and it's quite simple to understand, especially given that you left an explanatory comment. I'll merge now. Thanks 👍

Issue #13, opened 2016-12-02
Does not work on macOS (unless browser is started from terminal)


jceloria:
browserpass 0.1.2
Firefox: 50.0.2
Chrome: Version 54.0.2840.98 (64-bit)
macOS: 10.12.1-x86_64
```
$─► brew info pass
pass: stable 1.6.5 (bottled), HEAD
Password manager
https://www.passwordstore.org/
/usr/local/Cellar/pass/1.6.5_1 (27 files, 113.1K) *
  Poured from bottle on 2016-11-14 at 02:22:01
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pass.rb
==> Dependencies
Required: pwgen ✔, tree ✔, gnu-getopt ✔
==> Requirements
Required: gpg2 ✔
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completion has been installed to:
  /usr/local/share/zsh/site-functions
```
I installed the browser extensions from the Chrome webstore and Firefox, downloaded browserpass-darwinx64.zip and ran install.sh, selected both Chrome and Firefox and restarted both browsers. When I attempt to use browserpass I'm just presented with the message "Error talking to Pass"

I am able to run pass from the command line without issue:
```
$─► pass Games/ea.com | sed -e '1 s/^.*$/none-yo/g' -e 's/^\(login: \).*/\1 business/g'
none-yo
---
login:  business
url: https://www.ea.com
```

Am I missing something?

emersion:
Can you start Chrome from the command line and see if an error message is printed?

jceloria:
Sure, thank you for the response!

[28164:60163:1211/113143:ERROR:native_process_launcher.cc(137)] Can't find manifest for native messaging host com.dannyvankooten.browserpass

jceloria:
Sorry, I had forgotten that I removed the binary, I reinstalled and got the following:
LaunchProcess: failed to execvp:
/Users/jceloria/Applications/browserpass-darwinx64/browserpass-darwinx64

I then checked the path to the binary and it was not executable, I changed the permissions and things are working as expected for Chrome, thank you and my apologies for not catching something so simple.

jceloria:
I get the following with Firefox:

1481480385534	addons.webextension.browserpass@dannyvankooten.com	WARN	Loading extension 'browserpass@dannyvankooten.com': Reading manifest: Error processing author: An unexpected property was found in the WebExtension manifest.
[...]

Issue #23, opened 2016-12-26
Firefox plugin not prompting for gpg password


marceloneil:
browserpass: v1.0.1
browserpass plugin: v1.0.2
pass: v1.6.5
firefox: 50.1.0
linux: x86_64 Linux 4.8.13-1-ARCH

I've been using the chrome plugin with no problem, but I've been experiencing a slight issue with the firefox plugin. Whenever I try to use the plugin when the password for my gpg key has NOT been cached, the plugin immediately closes and I get this error. I am not prompted for the password like in chrome.

![screenshot from 2016-12-26 16-36-25](https://cloud.githubusercontent.com/assets/6353492/21486414/9d4bf808-cb8a-11e6-9fb9-1aeb532748e6.png)

When the password is cached (ie. I recently used the chrome plugin/pass), the firefox plugin works fine and enters in my password. I still get this output however:

![screenshot from 2016-12-26 16-38-26](https://cloud.githubusercontent.com/assets/6353492/21486431/e6b0daa4-cb8a-11e6-93af-8ce3762458aa.png)

Let me know if there is anything else I can do to resolve this bug.

Thanks,

Marcel

Timvde:
Same here, but on Nightly I get a different output in the browser console.

```
Cannot send message: Other side disconnected: ["MessageChannel:Response", {result:4, messageName:"20-0", recipient:{}, error:{}}]  ExtensionUtils.jsm:1091
	sendAsyncMessage resource://gre/modules/ExtensionUtils.jsm:1091:5
	_handleMessage/deferred.promise< resource://gre/modules/MessageChannel.jsm:671:9
stderr output from native app com.dannyvankooten.browserpass: 2017/01/06 11:28:19 exit status 2
```

DamienCassou:
I have the same problem.

kamidon:
I also have this problem. I think there must be something different about the environment available to the browserpass program when invoked by firefox vs. when invoked by chrome because I was able to confirm that if I run the browserpass command directly in a terminal and feed it the input it expects to get to be asked to retrieve a password, it *does* pop up the gpg-agent passphrase dialog and retrieve the password. Somehow the pass command when run by browserpass run by firefox isn't able to find the agent to trigger the passphrase request and thus the pass command fails and browserpass returns an error to firefox

I'm not sure what is different about the environment when firefox is running browserpass that triggers this nor whether there is anything that the extention author can do to fix it beyond failing a more specific bug with the firefox team.

[...]

PR #159, opened 2017-09-27
Find gpg even if it is not in $PATH, cleanup gpg options

maximbaz:
- Detect location of gpg binary even if $PATH is not set
- Cleanup gpg options
    - --use-agent is obsolete and has no effect since v2 according to man 
    - --batch existed even in v1, why not enable it always (and who the hell is using v1 today anyways?)

#13

--------

[browserpass-windows64.zip](https://github.com/dannyvankooten/browserpass/files/1338957/browserpass-windows64.zip)
[browserpass-freebsd64.zip](https://github.com/dannyvankooten/browserpass/files/1338958/browserpass-freebsd64.zip)
[browserpass-openbsd64.zip](https://github.com/dannyvankooten/browserpass/files/1338960/browserpass-openbsd64.zip)
[browserpass-darwinx64.zip](https://github.com/dannyvankooten/browserpass/files/1338959/browserpass-darwinx64.zip)
[browserpass-linux64.zip](https://github.com/dannyvankooten/browserpass/files/1338961/browserpass-linux64.zip)

### DIFF
--- a/browserpass.go
+++ b/browserpass.go
@@ -93,7 +93,7 @@ func Run(stdin io.Reader, stdout io.Writer, s pass.Store) error {
 	}
 }
 
-func detectGPGBin() (string, []string, error) {
+func detectGPGBin() (string, error) {
 	binPriorityList := []string{
 		"gpg2", "/bin/gpg2", "/usr/bin/gpg2", "/usr/local/bin/gpg2",
 		"gpg", "/bin/gpg", "/usr/bin/gpg", "/usr/local/bin/gpg",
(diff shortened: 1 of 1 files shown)


qbit:
I'd like to see PATH parsed first (maybe the first entry in binPriorityList being a split list of PATH if it is set?). Then fallback to guessing - reason being stuff like `pkgsrc` where install locations can vary pretty wildly will break (but likely the entries will be in PATH).

maximbaz:
There is no need to parse PATH, the first entry in binPriorityList is plain "gpg2", so just like it works today, it gpg2 is in PATH - it will be used. Same applies for gpg, as it is also present in binPriorityList.

qbit:
Excellent!

maximbaz:
Have confirmations that it helps with both #13 and #23, thus merging.

PR #161, opened 2017-09-28
Use browserify from node_modules

DamienCassou:
There is no reason to use a potentially absent one if there is always
one present in node_modules.


maximbaz:
Huh, indeed - thanks!

Issue #116, opened 2017-08-09
Dark theme


emersion:
Addon is unreadable when enabling dark theme.

How to reproduce : choose Customize > Themes > Compact dark

This is also the default Firefox Dev Edition theme.

![screenshot from 2017-08-09 15-40-29](https://user-images.githubusercontent.com/506932/29124447-39ec422c-7d19-11e7-8ea3-0d19627fc3fd.png)

00dani:
This issue isn't as bad under MacOS, which uses readable colours:

![screen shot 2017-09-21 at 08 48 36](https://user-images.githubusercontent.com/610876/30671324-aed34174-9ea9-11e7-818a-8c99ad9e7a92.png)

A dark theme would still be nice to have nonetheless though, since everything around it is dark. 😉

maximbaz:
Thanks to @aeris and #166 the colors are finally fixed when using dark theme. 

Implementing a dark theme for the extension is a different topic though, definitely out of scope of this ticket. 

I couldn't find anything that would allow detecting what color scheme the browser is using (does anyone know any other extension, that adapts its own colors based on the currently used browser theme?). 

Perhaps the only way is to add a theme selector to our own options page, and implement a second stylesheet for dark theme, and somehow dynamically inject in the popup one or another stylesheet based on the user preference.

If someone wants to continue discussion on this, please create a separate ticket (or better yet, a PR 😛).

PR #179, opened 2017-10-13
Better TOTP display

ashkitten:
It feels sorta bad injecting HTML this way from a script, but it works, right?

### DIFF
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -2,10 +2,13 @@
 
 var app = "com.dannyvankooten.browserpass";
 
+var tabLogins = {};
+
 chrome.runtime.onMessage.addListener(onMessage);
+chrome.tabs.onUpdated.addListener(onTabUpdated);
 
 // fill login form & submit
-function fillLoginForm(login) {
+function fillLoginForm(login, tabId) {
   const loginParam = JSON.stringify(login);
   const autoSubmitParam = JSON.stringify(localStorage.getItem("autoSubmit"));
 
@@ -23,17 +26,26 @@ function fillLoginForm(login) {
   );
 
   if (login.digits) {
-    chrome.tabs.executeScript(
-      {
-        file: "/inject_otp.js"
-      },
-      function() {
-        chrome.tabs.executeScript({
-          code: `browserpassDisplayOTP(${loginParam});`
-        });
-      }
-    );
+    tabLogins[tabId] = loginParam;
+    displayOTP(tabId);
   }
+}
+
+function displayOTP(tabId) {
+  chrome.tabs.executeScript(
+    tabId,
+    {
+      file: "/inject_otp.js"
+    },
+    function() {
+      chrome.tabs.executeScript(
+        tabId,
+        {
+          code: `browserpassDisplayOTP(${tabLogins[tabId]});`
+        }
+      );
+    }
+  );
 }
 
 function onMessage(request, sender, sendResponse) {
@@ -53,7 +65,7 @@ function onMessage(request, sender, sendResponse) {
         ) {
           // do not send login data to page if URL changed during search.
           if (tabs[0].url == request.urlDuringSearch) {
-            fillLoginForm(response, request.urlDuringSearch);
+            fillLoginForm(response, tabs[0].id);
           }
         });
 
@@ -63,5 +75,15 @@ function onMessage(request, sender, sendResponse) {
 
     // Need to return true if we are planning to sendResponse asynchronously
     return true;
+  }
+
+  if (request.action == "dismissOTP" && sender.tab.id in tabLogins) {
+    delete tabLogins[sender.tab.id];
+  }
+}
+
+function onTabUpdated(tabId, changeInfo, tab) {
+  if (changeInfo.status == "complete" && tabId in tabLogins) {
+    displayOTP(tabId);
   }
 }
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -2,13 +2,13 @@
 
 var app = "com.dannyvankooten.browserpass";
 
-var tabLogins = {};
+var tabInfos = {};
 
 chrome.runtime.onMessage.addListener(onMessage);
 chrome.tabs.onUpdated.addListener(onTabUpdated);
 
 // fill login form & submit
-function fillLoginForm(login, tabId) {
+function fillLoginForm(login, tab) {
   const loginParam = JSON.stringify(login);
   const autoSubmitParam = JSON.stringify(localStorage.getItem("autoSubmit"));
 
@@ -26,8 +26,11 @@ function fillLoginForm(login, tabId) {
   );
 
   if (login.digits) {
-    tabLogins[tabId] = loginParam;
-    displayOTP(tabId);
+    tabInfos[tab.id] = {
+      login: loginParam,
+      hostname: getHostname(tab.url)
+    };
+    displayOTP(tab.id);
   }
 }
 
@@ -41,7 +44,7 @@ function displayOTP(tabId) {
       chrome.tabs.executeScript(
         tabId,
         {
-          code: `browserpassDisplayOTP(${tabLogins[tabId]});`
+          code: `browserpassDisplayOTP(${tabInfos[tabId].login});`
         }
       );
     }
@@ -65,7 +68,7 @@ function onMessage(request, sender, sendResponse) {
         ) {
           // do not send login data to page if URL changed during search.
           if (tabs[0].url == request.urlDuringSearch) {
-            fillLoginForm(response, tabs[0].id);
+            fillLoginForm(response, tabs[0]);
           }
         });
 
@@ -77,13 +80,26 @@ function onMessage(request, sender, sendResponse) {
     return true;
   }
 
-  if (request.action == "dismissOTP" && sender.tab.id in tabLogins) {
-    delete tabLogins[sender.tab.id];
+  if (request.action == "dismissOTP" && sender.tab.id in tabInfos) {
+    delete tabInfos[sender.tab.id];
   }
 }
 
 function onTabUpdated(tabId, changeInfo, tab) {
-  if (changeInfo.status == "complete" && tabId in tabLogins) {
+  if (changeInfo.url && tabId in tabInfos) {
+    if (getHostname(changeInfo.url) != tabInfos[tabId].hostname) {
+      delete tabInfos[tabId];
+    }
+  }
+
+  if (changeInfo.status == "complete" && tabId in tabInfos) {
     displayOTP(tabId);
   }
+}
+
+function getHostname(url) {
+  // Manipulate the browser into parsing the URL for us
+  var a = document.createElement("a");
+  a.href = url;
+  return a.hostname;
 }
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -21,6 +21,19 @@ function fillLoginForm(login) {
       });
     }
   );
+
+  if (login.digits) {
+    chrome.tabs.executeScript(
+      {
+        file: "/inject_otp.js"
+      },
+      function() {
+        chrome.tabs.executeScript({
+          code: `browserpassDisplayOTP(${loginParam});`
+        });
+      }
+    );
+  }
 }
 
 function onMessage(request, sender, sendResponse) {
--- a/chrome/inject_otp.js
+++ b/chrome/inject_otp.js
@@ -13,6 +13,7 @@ window.browserpassDisplayOTP = function(login) {
     border-right: none;
     box-sizing: content-box;
     z-index: 1000000;
+    display: none;
   `;
 
   window.addEventListener("message", receiveMessage, false);
@@ -22,6 +23,8 @@ window.browserpassDisplayOTP = function(login) {
     }
 
     if (event.data.action == "resize") {
+      iframe.style.display = "block";
+
       iframe.width = event.data.payload.width;
       iframe.height = event.data.payload.height;
     }
--- a/chrome/inject_otp.js
+++ b/chrome/inject_otp.js
@@ -1,0 +1,36 @@
+window.browserpassDisplayOTP = function(login) {
+  iframe = document.createElement("iframe");
+  iframe.src = chrome.runtime.getURL("otp.html");
+  iframe.style = `
+    position: fixed;
+    top: 0;
+    right: 0;
+    background-color: white;
+    border-bottom-left-radius: 4px;
+    border-left: 1px solid #888888;
+    border-bottom: 1px solid #888888;
+    border-top: none;
+    border-right: none;
+    box-sizing: content-box;
+    z-index: 1000000;
+  `;
+
+  window.addEventListener("message", receiveMessage, false);
+  function receiveMessage(event) {
+    if (event.data.action == "load") {
+      iframe.contentWindow.postMessage(login, "*");
+    }
+
+    if (event.data.action == "resize") {
+      iframe.width = event.data.payload.width;
+      iframe.height = event.data.payload.height;
+    }
+
+    if (event.data.action == "dismiss") {
+      document.body.removeChild(iframe);
+      window.removeEventListener("message", receiveMessage);
+    }
+  };
+
+  document.body.appendChild(iframe);
+}
--- a/chrome/inject_otp.js
+++ b/chrome/inject_otp.js
@@ -1,5 +1,6 @@
 window.browserpassDisplayOTP = function(login) {
   iframe = document.createElement("iframe");
+  iframe.id = "browserpass-otp-iframe";
   iframe.src = chrome.runtime.getURL("otp.html");
   iframe.style = `
     position: fixed;
@@ -30,10 +31,14 @@ window.browserpassDisplayOTP = function(login) {
     }
 
     if (event.data.action == "dismiss") {
-      document.body.removeChild(iframe);
+      iframe.remove();
       window.removeEventListener("message", receiveMessage);
     }
   };
 
+  var oldIframe = document.getElementById("browserpass-otp-iframe")
+  if (oldIframe != null) {
+    oldIframe.remove();
+  }
   document.body.appendChild(iframe);
 }
--- a/chrome/inject_otp.js
+++ b/chrome/inject_otp.js
@@ -2,6 +2,7 @@ window.browserpassDisplayOTP = function(login) {
   iframe = document.createElement("iframe");
   iframe.id = "browserpass-otp-iframe";
   iframe.src = chrome.runtime.getURL("otp.html");
+  iframe.scrolling = "no";
   iframe.style = `
     position: fixed;
     top: 0;
@@ -14,6 +15,7 @@ window.browserpassDisplayOTP = function(login) {
     border-right: none;
     box-sizing: content-box;
     z-index: 1000000;
+    overflow: hidden;
     visibility: hidden;
   `;
 
--- a/chrome/inject_otp.js
+++ b/chrome/inject_otp.js
@@ -14,7 +14,7 @@ window.browserpassDisplayOTP = function(login) {
     border-right: none;
     box-sizing: content-box;
     z-index: 1000000;
-    display: none;
+    visibility: hidden;
   `;
 
   window.addEventListener("message", receiveMessage, false);
@@ -24,7 +24,7 @@ window.browserpassDisplayOTP = function(login) {
     }
 
     if (event.data.action == "resize") {
-      iframe.style.display = "block";
+      iframe.style.visibility = "visible";
 
       iframe.width = event.data.payload.width;
       iframe.height = event.data.payload.height;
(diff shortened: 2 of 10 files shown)


ashkitten:
er, scratch the "works" part... it doesn't persist to the next page where you actually have to *enter* the TOTP

chrome/background.js
```diff
@@ -21,6 +21,17 @@ function fillLoginForm(login) {
       });
     }
   );
+
+  chrome.tabs.executeScript(
+    {
+      file: "/injecttop.js"
```
maximbaz: For clarify let's be specific and just call this file `inject_totp.js` or something like that - we can always rename if there will be another use for this file later.

chrome/background.js
```diff
@@ -21,6 +21,17 @@ function fillLoginForm(login) {
       });
     }
   );
+
+  chrome.tabs.executeScript(
```
maximbaz: could you wrap this in an `if` so that this injection happens if and only if the `login` object actually contains an OTP token?

chrome/injecttop.browserify.js
```diff
@@ -0,0 +1,49 @@
+window.browserpassDisplayTOTP = function(login) {
+  if(login.digits) {
```
maximbaz: ... and thus remove this check here 🙂

makefile
```diff
@@ -29,6 +29,9 @@ chrome/script.js: chrome/script.browserify.js
 chrome/inject.js: chrome/inject.browserify.js
 	$(BROWSERIFY) chrome/inject.browserify.js -o chrome/inject.js
 
+chrome/injecttop.js: chrome/injecttop.browserify.js
+	$(BROWSERIFY) chrome/injecttop.browserify.js -o chrome/injecttop.js
```
maximbaz: I need to read myself as well what exactly this `browserify` does, but looking at the currently present examples, we name the file `something.browserify.js` and gitignore `something.js`, and this command will generate this `something.js` during the build.

If this `browserify` action is also needed in your case, you should rename your file to `inject_totp.browserify.js` or something like that, and add a `.gitignore` entry for `inject_totp.js`.

chrome/injecttop.browserify.js
```diff
@@ -0,0 +1,49 @@
...
+    var totpBox = document.createElement("div");
+
+    totpBox.style = `
```
maximbaz: I found [chrome.tabs.insertCSS](https://developer.chrome.com/extensions/tabs#method-insertCSS), I think we should move all the styles in a some kind of `otp.css` and inject that file in the page.

chrome/injecttop.browserify.js
```diff
@@ -0,0 +1,49 @@
...
+      border-left: 1px solid #888888;
+      border-bottom: 1px solid #888888;
+      font-family: monospace;
+      z-index: 1000000;
+    `;
+
+    var totpText = document.createElement("span");
+    totpBox.appendChild(totpText);
+
+    totpText.innerText = "TOTP:";
```
maximbaz: Previously it was `(login.label || "OTP") + ":"`, I think you should preserve that 🙂

maximbaz:
Awesome job! I left some notes, mostly about code style. Preserving the box between changing pages sounds like a tough problem, actually... Do you have some ideas on how to solve it?

@qbit FYI the initial PR is available, subscribe 🙂

ashkitten:
I'm thinking I should actually use an iframe to embed the TOTP box, since then it won't be affected by the page's styles or scripts.

maximbaz:
Oh yeah, that's a good point

ashkitten:
And then we can just use an HTML document instead of injecting it into the page, as well

makefile
maximbaz: `browserify` is needed only if you have `require()` calls in your code. You don't, so you don't need to use browserify, and you can remove these 2 lines from `makefile` - they are not needed. Just name your file without `browserify` suffix and everything will be good.

The fact that `inject.js` is called `inject.browserify.js` and has the similar entry in `makefile` is just a mistake, it doesn't need to be declared like so, as we don't `require()` anything in `inject.js`. I'll deal with it separately.

ashkitten:
Hmm, the problem with the iframe approach is that it's not possible to size the iframe to fit its content, so we're stuck with a fixed size element.

ashkitten:
We might be able to work around this limitation by using the script that's injected into all frames and communicate via the background script when we need to resize, but that feels incredibly hacky.

maximbaz:
Let's do a fixed-size approach, and make sure it fits most of the OTP tokens - what is the maximal length that you've seen? I've seen 7 digits, so if we can make it look good for 10 digits (just in case), it should be enough 🙂

qbit:
Max size is 11 : https://github.com/gokyle/twofactor/blob/master/oath.go#L101

qbit:
the file name should be "inject_otp" - since it is technically doing either TOTP or HOTP - depending on what the url is configured for. Also it seems that this breaks the makefile (for at least `release`):

```
dep ensure
./node_modules/.bin/browserify chrome/script.browserify.js -o chrome/script.js
./node_modules/.bin/browserify chrome/inject.browserify.js -o chrome/inject.js
gmake: *** No rule to make target 'chrome/injecttop.browserify.js', needed by 'chrome/injecttop.js'.  Stop.
```

ashkitten:
By the way, why does the .editorconfig specify tabs for all files except .js? Except that none of the files use tabs?

maximbaz:
Emm... I came to the project after it was created and honestly I haven't simply noticed that file until now 🙂 

Go files use tabs, and javascript files use spaces - I think that was the intention.

My editor automatically formats go code with `gofmt` and javascript code with `prettier`, but I don't think it reads `.editorconfig` to be honest.

ashkitten:
@maximbaz @qbit this latest commit switches to using an iframe. I was able to get automatic sizing to work via `Window.postMessage()`. It uses three new files - `otp.html`, `otp.css`, and `otp.js` which are only for the iframe.

qbit:
@ashkitten do you have any OTP sites that reload after entering the creds? The OTP display vanishes before you can even read what it is displaying. A few examples are lobste.rs, gandi.net and mastodon.social

ashkitten:
gitlab.com does as well. Some websites like tumblr.com and google will also change pages before allowing you to enter your password, so we should figure that out as well.

qbit:
Maybe a delay can be introduced (configurable length via the options page) before submitting? Also on the subject of config stuffs, i think it would be good to be able to disable otp all together in the options :D

ashkitten:
I don't think a delay should be the answer. I think the solution is to have the otp iframe persist between pages, or at least be recreated.

maximbaz:
> Some websites like tumblr.com and google will also change pages before allowing you to enter your password, so we should figure that out as well.

We just treat such case as two different pages. If there is only one field (email or password), browserpass will only fill out that single field. When you are redirected to a next page also having one field, fire browserpass again. Works well on google today, I don't think we should introduce any complications to this. 

But OTP does seem to have to be persistent across page changes.

> I think it would be good to be able to disable otp all together in the options.

Don't think that's necessary... If you are storing otp tokens in your pass, it means you want to use them. If you don't have them in your pass, the `otp.js` will not even be injected in the page.

chrome/otp.js
```diff
@@ -0,0 +1,30 @@
+var totpInput = document.getElementById("totp");
```
maximbaz: Let's use `otp` instead of `totp` everywhere, as it was suggested that it could be `hotp` as well.

makefile
```diff
@@ -29,6 +29,9 @@ chrome/script.js: chrome/script.browserify.js
 chrome/inject.js: chrome/inject.browserify.js
 	$(BROWSERIFY) chrome/inject.browserify.js -o chrome/inject.js
 
+chrome/inject_otp.js: chrome/inject_otp.browserify.js
+	$(BROWSERIFY) chrome/inject_otp.browserify.js -o chrome/inject_otp.js
+
```
maximbaz: This comment was probably lost in the history: because `inject_otp.js` doesn't have `require()` calls, it doesn't need to be browserified. Just remove this piece, rename the file to `inject_otp.js` and remove this entry from `.gitignore` - I'll take care of doing the same for `inject.browserify.js` separately later today.

ashkitten: Sounds good, I was just following what had been done in the past.

chrome/otp.js
ashkitten: Looks like I forgot about that. Will fix.

maximbaz:
So the only unsolved problem is how to persist/recreate the iframe, right? Do you have any ideas how this can be implemented?

ashkitten:
The only way I can think of to do that is to tell the background script when the otp is dismissed, and keep injecting inject_otp.js until it gets dismissed

maximbaz:
I guess it's a valid idea, just need to be aware that background script is one for all tabs, so it's important not to begin inject iframe in every single opened tab 🙂

ashkitten:
Actually, is there any way to keep the popup open while the user does an action on the page?

maximbaz:
No, this is a limitation of the browser that lead us to create background page in the first place - as soon as popup loses focus, it gets automatically killed by the browser.

ashkitten:
yeah, that's what I thought...

maximbaz:
I've unbrowserified inject.js, you'll have a conflict in that file because it got renamed from `inject.browserify.js` to just `inject.js`.

ashkitten:
Fixed merge conflicts and squashed all my commits

maximbaz:
Cool! Just want to confirm, you want to implement continuous injection of the iframe before this is merged, right?

ashkitten:
That would be optimal, but if you're comfortable with merging now and fixing that in a different PR I'd be okay with that as well.

maximbaz:
There's no rush, I can merge at any time, then let's make it here and merge a complete package at once.

ashkitten:
@maximbaz This last commit adds persistent OTP display by adding a listener for chrome.tabs.onUpdated and storing logins in an object with the tab ID as a key until dismissed, upon which the login is deleted from the object.

chrome/background.js
```diff
@@ -21,6 +24,23 @@ function fillLoginForm(login) {
       });
     }
   );
+
+  if (login.digits) {
+    chrome.tabs.query({ active: true }, function(tabs) {
```
maximbaz: this is already queried in `onMessage` where `fillLoginForm` is called, maybe just pass the tabID from there and don't query again?

chrome/background.js
```diff
@@ -51,4 +71,27 @@ function onMessage(request, sender, sendResponse) {
...
   }
+
+  if (request.action == "dismissOTP" && sender.tab.id in tabLogins) {
+    delete tabLogins[sender.tab.id];
+  }
+}
+
+function onTabUpdated(tabId, changeInfo, tab) {
+  if (changeInfo.status == "complete" && tabId in tabLogins) {
+    chrome.tabs.executeScript(
```
maximbaz: could you please extract this `chrome.tabs.executeScript` call to a separate method and call it here and in another place? In `fillLoginForm` you can have tabId too, so that shouldn't be a problem.

maximbaz:
Looking good! I left a couple of final comments, please have a look and in meantime I'll configure an otp entry for myself and give it a test 🙂

@qbit please test as well!

ashkitten:
Done and done!

maximbaz:
I'm testing now, I like the simplicity of the popup, looks quite neat!

I'm not sure how I feel about resizing, it is visible when the size changes and it is a little bit annoying... Do you also see this? Are you sure you don't want to just fix the size to fit 10 digits?

maximbaz:
Functionality-wise, the most obvious next improvement would be to have the OTP code refreshing and have a count-down like other authenticators usually do... I think you already mentioned this idea earlier. But I recommend taking that separately.

ashkitten:
The problem I realized with a fixed size is that we have a label which is not a fixed size. I would be fine with setting the label to just say `OTP:` which would make it more possible.

maximbaz:
That is a good point. I would be fine with this too, but let's wait for @qbit as it was him who introduced showing custom label instead of a constant `OTP:`. @qbit, what do you want more, non-jumping iframe or custom label? 🙂

maximbaz:
I'm not sure how big of a problem this is, but I noticed that if I open website X, fill in credentials to get OTP popup visible, don't click "Done" but instead type a completely different url in that tab and move to website Y, the popup will be there as well.

I guess that's okay? I'm thinking that we can't just remember the old url, it might change between page with credentials and page with otp token request.

Another idea on a related note, maybe we should replace the button "Done" with "Copy", which will copy the code to clipboard and close the popup, and expect that most people will do just that?

ashkitten:
I think it's fine to have the pop-up persist between URLs. If you want I can make it only persist on the same domain.

As for replacing the "Done" button, I think it should stay. It's not very obvious how to close the dialog otherwise. Maybe we can turn it into a smaller, X button?

maximbaz:
I think it would be sensible to get rid of popup once domain changes... 🤔 Let's do it.

I was thinking more like... who would choose to manually copy the digits it there is a "Copy" button that can do the same action with fewer clicks (it will copy the digits and close dialog). But you are right that if for whatever reason I decided to manually select the text and press <kbd>Ctrl+C</kbd>, then I would be confused about how to close this dialog.

Do you like the general idea about this Copy button, i.e. would you use it? If yes, I think we should have both "Copy" and a small X button, if not, the current "Done" is enough.

ashkitten:
For the resizing issue, we can also just hide the iframe until it's ready

ashkitten:
The copy button would be fine, I can add that as well.

maximbaz:
Awesome idea for fixing the resize issue!

Then let's do it like that and that would be it, I don't see any other issues with the code and how it behaves.

- [ ] Remove popup when domain changes (subdomain? top level domain?)
- [ ] Hide iframe until its size is known
- [ ] Add Copy button and a small X

ashkitten:
I'd like to make the copy button an icon, where should I put such an icon?

maximbaz:
The `chrome` folder has a few icons already, put it in there as well

ashkitten:
kapow! code!

I didn't do any button styling, just what's necessary. If anyone feels strongly about this they can make their own PR.

qbit:
Nice! Next issue seems to be the OTP window staying open on subsequent clicks within the same domain - even after closing with the X.  :D

chrome/background.js
```diff
@@ -51,4 +79,27 @@ function onMessage(request, sender, sendResponse) {
...
   }
+
+  if (request.action == "dismissOTP" && sender.tab.id in tabInfos) {
+    delete tabInfos[sender.tab.id];
+  }
+}
+
+function onTabUpdated(tabId, changeInfo, tab) {
+  if (changeInfo.url) {
+    if (getHostname(changeInfo.url) != tabInfos[tabId].hostname) {
```
maximbaz: I've got an error on this line in the console: `Error in event handler for tabs.onUpdated: TypeError: Cannot read property 'hostname' of undefined`

chrome/background.js
```diff
@@ -51,4 +79,27 @@ function onMessage(request, sender, sendResponse) {
...
+  if (changeInfo.status == "complete" && tabId in tabInfos) {
+    displayOTP(tabId);
+  }
+}
+
+function getHostname(url) {
+  // Manipulate the browser into parsing the URL for us
+  var a = document.createElement("a");
+  a.href = url;
+  return a.hostname;
```
maximbaz: haha, I like this approach :smile:  :+1:

ashkitten:
To address the dismiss button not working properly now: for some reason onMessage is not receiving the message from otp.js... Not sure why?

---

Edit: apparently switching the order of the messages sent in otp.js dismiss onclick handler fixes it??? weird... I can't figure out why this is happening

chrome/background.js
```diff
@@ -21,9 +24,36 @@ function fillLoginForm(login) {
...
+        {
+          code: `browserpassDisplayOTP(${tabInfos[tabId].login});`
+        }
+      );
+    }
+  );
 }
 
 function onMessage(request, sender, sendResponse) {
+  console.log(JSON.stringify(request));
```
maximbaz: need to remember to delete this before merging

ashkitten: Oops lol, thanks for reminding me :)

maximbaz:
Maybe message execution is synchronous (although it shouldn't be), and thus if `dismiss` message goes first, it removes the iframe without giving it a chance to send `dismissOTP` message to the background page... The good thing is that it works 🙂

Now it looks all good, @qbit what do you say?

But I've got some bad news, I've decided to try this on Firefox before merging, and it doesn't work there at all 😞 Looks like the messages are not being sent/received, I see an invisible iframe on the page with no content 😞

chrome/inject_otp.js
```diff
@@ -0,0 +1,39 @@
+window.browserpassDisplayOTP = function(login) {
+  iframe = document.createElement("iframe");
```
maximbaz: One tiny issue: if I run browserpass twice on the same page, it will create two iframes. Then I will never be able to close the one that was opened first.

Could you please assign a unique ID to this iframe (e.g. `browserpass-otp-iframe`), and on the first line here delete an existing element with the same id? `document.getElementById("browserpass-otp-iframe").remove()` should do it I think.

ashkitten: Done

maximbaz: Yup, tested, thanks! Now only FF issue remains, I hope it's something we could easily workaround, but I can't figure out so far what exactly is missing...

ashkitten: I don't use Firefox, so if either of you are able to figure out the issue that would be fantastic.

maximbaz: I don't use it too, but I had to install it to test PRs for this extension from time to time 😄 

So far I can only see that `receiveMessage()` in `inject_otp.js` is never being called, and I checked that according to docs the `postMessage` should be supported since forever. There are no errors in the console 😞

maximbaz:
#### Firefox issue:

One finding - you did a change in `chrome/manifest.json`, do the same in `firefox/manifest.json` 🙂
That will make it load, at least. 

However, it doesn't seem to always load, if I put breakpoint in `receiveMessage()` in `inject_otp.js` and go through the function step by step - then it will work. If I run browserpass with devtools closed - then there is no popup.

And when popup does load, it looks like this:

image

I'm afraid I won't be able to fix all the issues myself, either you'll have to install Firefox or we can put it on hold until someone comes to help with Firefox part - it would be wrong to merge this and break existing functionality for all FF users 😞

ashkitten:
So sounds like a race condition if it's working when stepping through. I'll get firefox running and work on it in a bit.

maximbaz:
Sounds like so... This is what it looks like when there is no popup:

image

maximbaz:
A hint on how to load the extension in Firefox: 
open `about:debugging#addons`  and use `Load temporary Add-on`.

ashkitten:
The reason it's setting height and width to 0 is because in Firefox, `visibility: hidden` apparently means it doesn't take up any space either, so it returns 0

ashkitten:
@maximbaz should work now

ashkitten:
hmm, the `Tab.onUpdated` listener's `changeInfo` object doesn't seem to be containing any `url` keys

maximbaz:
Found the issue - there are missing permissions in the `firefox/manifest.json`, compare them to the `chrome/manifest.json` - could you add them please? I'm surprised that nobody saw this earlier 😄

maximbaz:
Styling request, I know this changes look in Chrome, but not to re-introduce issue #116, please add:

```
#otp {
  background: white;
  color: black;
}
button {
  background: #f1f1f1;
  color: black;
}
```

Before:

image


After:

image

ashkitten:
Alright, I added some styles for the buttons so they look halfway decent.

Good now? ;)

maximbaz:
Your styling is much better that what I offered 😃 

But it broke in Chrome, sorry... :man_facepalming: 

image

In FF the popup persists across changing domains...

ashkitten:
sigh... everything i do makes more bugssss

hmm, I'm not getting that in chromium, what's your deal?

maximbaz:
We are soooo close! I want to merge it today :D 

I'm on Chromium on Linux, I'll try to see myself which style change did that...

ashkitten:
Like I said, it's not seeming to be triggering `onTabUpdated` with a `url` key in the changeInfo on Firefox. I'm not sure why this is.

maximbaz:
Yeah, I came to that conclusion just now too. I thought that bringing permissions to firefox would fix the missing `url` property, but apparently this is not true. Let's leave it like that then.

ashkitten:
Ah, it requires the "tabs" permission to show the url in firefox.

ashkitten:
I'll add the tabs permission to both the manifests just to be safe

maximbaz:
Okay, thanks.

I found a solution to the scrollbars. Add this to `inject_otp.js`:

```
iframe.scrolling = "no";
iframe.style = `
  ...
  overflow: hidden;
`;
```

ashkitten:
Done :)

maximbaz:
Thanks, doing final testing now!

maximbaz:
Everything seems to be working perfectly good 🎉 
Thanks so much for getting it to the end, this has been quite a change!

Merging now and releasing a new version.

Issue #105, opened 2017-07-07
option to copy username/password to clipboard.


james-lawrence:
for pages that fail to autofill a nice feature would be to allow copying fields to the clipboard.

bddenhartog:
Personally, I'd like to see this implemented either with an icon/button to click (maybe the favicon? it could be replaced with a copy icon on row hover), or simply via right-clicking the row. Anything else, IMHO, would complicate the workflow of getting the password beyond alt+tabbing to a terminal and typing out the `pass` command.

maximbaz:
Right-clicking on a row is not intuitive, I'm afraid nobody will guess that such functionality exists 🙂

My suggestions:
- Have two buttons on the right side of every entry, one to copy username, another to copy password.
- Handle <kbd>Ctrl+C</kbd> (is it <kbd>Cmd+C</kbd> on Mac?) and copy the password of the currently focused entry (this item is for keyboard users).

bddenhartog:
> Right-clicking on a row is not intuitive

That's most certainly a subjective statement. Right-clicking was the first thing I did when imagining how a "copy" function would work - users are already used to interacting with the plugin with their mouse, wouldn't a keyboard shortcut detract from the experience they're used to?

maximbaz:
We have a very powerful support for keyboard users, definitely not everyone is interacting with the plugin with the mouse 🙂

That's why I want to see this feature implemented in two ways, for mouse users and for keyboard users. We can discuss how to best implement it for mouse users, we can discuss how to best implement it for keyboard users, but whatever we choose for "mouse approach" will not substitute the "keyboard approach". 

Now, for those who use mouse, we have two ideas now - right click on an entry, or left-clicking on one of two new buttons. I vote for the buttons because it is more intuitive, a random user who sees such buttons can immediately understand what they do. This is very approximate on how I imagine it:

![browserpass_icons](https://user-images.githubusercontent.com/1177900/31216018-4203976a-a9b2-11e7-81d8-0601daa42009.png)

For those who use keyboard, <kbd>Ctrl+C</kbd> (<kbd>Cmd+C</kbd> on Mac?) is the only idea that we have so far, and seems quite intuitive to me 🙂

Weker01:
Is someone working on this?

maximbaz:
Don't think so, wanna give it a try? 😉

Weker01:
Time to dive into a Codebase I know nothing about. I will try implementing something on the chrome side first.

[...]

PR #188, opened 2017-11-10
Copy to Clipboard

Weker01:
Tested only on Linux Chrome.

Should there be keybindings to focus the copybuttons?

---
Reference: #105


Weker01:
Just tested: Does not work in Firefox.

chrome/script.browserify.js
```diff
@@ -159,16 +173,67 @@ function getLoginData() {
   );
 }
 
+function usernameToClipboard() {
+  var username = this.split("/");
```
maximbaz: I don't store username as a file name, username should come from the host app, just like you get password below.

maximbaz:
Nice job 🙂

I'll be testing and recording some TODOs and general thoughts for us to tackle.

- [ ] The icons.... I'll try to search for something more beautiful 😃 
- [ ] When I press <kbd>Tab</kbd> to scroll through the list, the focus should never go to the username/password buttons, it should only stay on the line itself. In other words, if I have 3 items in the list, pressing <kbd>Tab</kbd> 3 times should go through all entries. Now it requires pressing <kbd>Tab</kbd> 9 times, which is just too much.
- [ ] Instead, we can try to implement a support for <kbd>Ctrl+C</kbd>, when a line is selected in the popup, pressing <kbd>Ctrl+C</kbd> should copy the password.
- [ ] Firefox support, you are right, the file is being decrypted, but somehow the contents is not being pushed to clipboard.

chrome/script.browserify.js
```diff
@@ -159,16 +173,67 @@ function getLoginData() {
...
+            window.close();
+          }
+        }
+      );
+    }
+  )
+}
+
+function toClipboard(s){
+  m.render(
```
maximbaz: I would just use plain javascript here, document.createElement - I know from other ticket that firefox treats this `m()` differently than chrome, it doesn't create the element immediately or something like that, maybe that's the reason why it doesn't work in FF

maximbaz:
Regarding icons, how about the following:

- reuse the existing `icon-key.svg` for the password
- take the attached file and save it as `icon-user.svg` for username (don't name it `icon-user-copy`, we might reuse it for something else later).

image

[icon-user.svg.zip](https://github.com/dannyvankooten/browserpass/files/1462298/icon-user.svg.zip)

What do you think?

Weker01:
It looks good in this example, but what if there is no fav. icon?

maximbaz:
Very good point - why don't we remove the icon from the left side entirely if there is no favicon? I don't see much point in it anyway, but having two keys on the left and on the right is confusing indeed, both if they are identical icons or different icons of a key 🙂

Let me know if you like this.

image

You don't have to fix it here, I can do it myself on master and you can then later rebase.

Weker01:
I am sorry. I can't work on this till next week. Life can be stressful...

maximbaz:
No worries at all, take your time! :+1:

Weker01:
Should now work in Firefox. funnily enough, the permission to write to the clipboard where just missing. One line fix.

maximbaz:
That's awesome! I confirm that it works in Firefox now. I'll go ahead and deal with the case of missing favicon, I'll ping you when it's ready so you can rebase. When you have time, can you please focus on the keyboard navigation next? I noticed that arrow navigation and <kbd>Tab</kbd> / <kbd>Shift+Tab</kbd> navigation has different behavior now, both have some issues:

- <kbd>Tab</kbd> goes through copy buttons, it shouldn't, instead it should only go through lines like arrows do now.
- arrows on the other side stopped cycling, previously you could press <kbd>Down</kbd> many times to go down through the list, but when you are at the bottom, next <kbd>Down</kbd> jumps back to the input box - now it doesn't do that, it gets stuck on the last entry.

chrome/script.browserify.js
maximbaz: even though it works in FF, I still think it's better to use vanilla js in this case

chrome/styles.css
```diff
@@ -33,16 +33,35 @@ body{
...
+.entry {
+  display:flex;
+  flex-flow: row;
+  border: 0;
+  border-bottom: 1px dotted #ccc;
+}
+
+.copy {
+  width: 32px;
+  border: 0;
```
maximbaz: please add `cursor: pointer;` to these buttons 🙂

Weker01:
Tab navigation is still a mess I agree.

maximbaz:
Can't claim to be an expert in this, but I think if you just add html attribute `tabindex="-1"` to the buttons, then they will be skipped by the tab navigation. Just something to try 🙂

maximbaz:
The change is on master now, sorry it introduced some conflicts in your branch - please rebase 🙂

Weker01:
Added Shift-C to copy the username. In my experience I need that pretty often.

maximbaz:
Awesome, I'm huge fan of shortcuts! 👍 Shift-C is a good choice.

maximbaz:
Tab navigation seems also be good now, only arrow navigation needs to match its exact behavior now, i.e. avoid being "stuck".

chrome/script.browserify.js
```diff
@@ -160,16 +180,73 @@ function getLoginData() {
...
+      if(chrome.runtime.lastError) {
+        error = chrome.runtime.lastError.message;
+        m.redraw();
+      }
+
+      chrome.tabs.query(
+        {lastFocusedWindow: true, active: true},
+        function(tabs) {
+          //do not copy login data if URL changed during search.
+          if(tabs[0].url == urlDuringSearch) {
```
maximbaz: we probably don't care if URL changed, we have this check for filling a form to avoid filling credentials on a wrong website, but putting decrypted username/password to clipboard is an action that is safe to do regardless of the currently open tab.

chrome/script.browserify.js
```diff
@@ -160,16 +180,73 @@ function getLoginData() {
...
 
+function loginToClipboard() {
+  var what = this.what;
+  chrome.runtime.sendNativeMessage(
+    app,
+    {action: "get", entry: this.entry},
+    function(response) {
+      if(chrome.runtime.lastError) {
+        error = chrome.runtime.lastError.message;
+        m.redraw();
```
maximbaz: need to make sure `window.close()` is not executed in this case, so users can see an error message

maximbaz:
@Weker01 how is your availability these days? I'd like to merge this today-tomorrow and make a new release. I can have a look at the last open points (arrow navigation, code comments) if you are busy.

chrome/script.browserify.js
Weker01: While I agree that it is not really needed, I still thing that we shouldn't write passwords to the clipboard when the site changed, so the user doesn't need to check check if it changed.

Weker01:
I should have some time to clean up today.

Weker01:
So whats left on the to-do?

chrome/script.browserify.js
maximbaz: I've just realized the following: we do similar url check in the background script, because background script is shared among all tabs, so there is a risk where you select credentials to be filled, but then quickly change tab in the browser. However this script is page-specific, so for example gpg is very slow and decryption takes seconds, if you change to a different tab during this time, the popup will close and the password will not be put in the clipboard. This can be reproducible by wrapping this whole `chrome.tabs.query` in a `setTimeout(..., 2000)` block.

Can you come up with an example when this check would still be useful? I'm fine with keeping it, but I want to know some examples of why we have it 🙂

maximbaz:
My question about the example when we need the url check, and that's all 🙂 I just tested arrow navigation, works perfectly! I'm still curious about the possibility to use vanilla js in `toClipboard`, I have a feeling that it will get simpler code. But maybe I'm wrong 🙂 It's not critical and I'll play with it on my own.

Awesome job on this PR!

chrome/script.browserify.js
Weker01: Ok, I didn't realize the script halts execution when one navigates. Yes you are right it's not needed.

maximbaz: thanks, merging now 👍

PR #189, opened 2017-11-11
More explicit signatory during verification.

johnmee:
Would it be a bad idea to use a trusted signature as well?


maximbaz:
I adjusted the format a bit, is that okay?

Could you explain what you mean by a trusted signature?

maximbaz:
I'm going to merge this as it is, please let me know if you have some further feedback / question and we can adapt 😉 Thanks for the idea!

PR #190, opened 2017-11-15
Tweak install.sh for Homebrew

zsau:


install.sh
```diff
@@ -120,27 +121,27 @@ ESCAPED_HOST_FILE=${HOST_FILE////\\/}
...
     echo "If you are running './install.sh' from the source code, make sure to follow CONTRIBUTING.md on how to build first."
     exit 1
   fi
-  cp "$DIR/firefox-host.json" "$TARGET_DIR/$APP_NAME.json"
+  cp "$JSON_DIR/firefox-host.json" "$TARGET_DIR/$APP_NAME.json"
 fi
 
 # Replace path to host
-sed -i -e "s/%%replace%%/$ESCAPED_HOST_FILE/" "$TARGET_DIR/$APP_NAME.json"
+sed -i '' -e "s/%%replace%%/$ESCAPED_HOST_FILE/" "$TARGET_DIR/$APP_NAME.json"
```
maximbaz: Unfortunately this doesn't work on linux 😞 Is there a way to make it compatible?

```
sed: can't read : No such file or directory
```

zsau: Whoops, didn't realize that was a sed BSD-ism. Perhaps instead of first copying the files and then using `-i`, we could just run sed on the original location and output to `NativeMessagingHosts`?

maximbaz: sounds like a good approach

maximbaz:
awesome, thanks :+1:

Issue #192, opened 2017-11-25
Running on Windows through WSL


jtsymon:
[This setup probably isn't worth supporting, I'm just leaving this here as a guide for anyone else with a similar setup]

I use browserpass on Linux, and occasionally have to use Windows, so I was pleased to find browserpass supports Windows... except I already use pass under WSL, which the Windows host doesn't work with. So I figured instead of setting up a second copy of my password store for the Windows host, I would just make browserpass go through WSL. Turns out to be quite simple:

1. Install the Windows host application, as well as the Linux host application (under WSL).
2. Create `%localappdata%\browserpass\browserpass-linux.bat` with the following contents:
```
@echo off
bash -c ~/browserpass/browserpass-linux64
```
3. Change the path in `%localappdata%\browserpass\browserpass-firefox.json` (or `-chrome.json`) to point to `browserpass-linux.bat`

Now the browser extension should be able to load password entries, but if you have a password on your GPG key it will probably fail to decrypt passwords due to the key not being loaded in your gpg-agent. Normally on Linux gpg-agent would just pop up a password input, but this doesn't work with WSL (maybe it is possible if you install X11 under WSL?)
Anyway, you can work around this by loading the GPG key into your gpg-agent inside WSL (e.g. just load a password inside WSL). Kind of annoying, but it works.

maximbaz:
Hey, thanks for sharing! I think these instructions are useful, but it would be difficult to find them among issues (I'd also like to keep only actionable issues open). Would you consider opening a PR and adding maybe a condensed description of this to the README.md? I suggest adding this to Installation section, perhaps as a new subsection "Installing the host application on Windows through WSL" after the subsection "Installing the host application".

frebib:
I'd like to add that this can be fully automated with a self-starting gpg-agent and Windows-invoked graphical password entry, similar to standard `pinentry` applications.

I managed to get this working with a combination of [https://github.com/diablodale/pinentry-wsl-ps1](pinentry-wsl-ps1) which uses Powershell and several hours of pulling my hair out. I have some reasonably complex bash and zsh configuration so this may need tweaking to work for everyone but here is roughly what I did:

- Download and locate the `pinentry-wsl-ps1` script (I put it in `/usr/local/bin/pinentry-wsl`)
- Configure `gpg-agent` to use the above pinentry application:
  - ```diff
     ~/.gnupg/gpg-agent.conf
     ------------------------------
[...]

PR #193, opened 2017-11-25
Add instructions for installing on Windows through WSL

jtsymon:
Moving instructions for working with WSL to README.md.
Some stuff to look at:
- Automatically create `browserpass-linux.bat` with `install.ps1` to simplify the instructions? (it can just sit there unused normally)
- If your GPG key has a password (which it really should), the host running under WSL won't be able to unlock it since it can't prompt the user for the password. This means you can't get decrypt any passwords unless you've already got the key loaded in gpg-agent.
Could possibly handle this by prompting for the master password in the browser extension in this case? Seems kinda sketchy though, especially for such an edge case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,19 @@ Installing the binary & registering it with your browser through the installatio
 
 #### Installing the host application on Windows through WSL
 
+If you already use `pass` under WSL and prefer to have a single copy of your password store, you can use browserpass through WSL as well.
 1. Install the Windows host application (see previous section) as well as the Linux host application (under WSL).
-2. Create `%localappdata%\browserpass\browserpass-linux.bat` with the following contents:
+2. Create `%localappdata%\browserpass\browserpass-wsl.bat` with the following contents:
 ```
 @echo off
-bash -c ~/browserpass/browserpass-linux64
+bash -c ~/.browserpass/browserpass-linux64
 ```
-3. Change the path in `%localappdata%\browserpass\browserpass-firefox.json` (or `-chrome.json`) to point to `browserpass-linux.bat`
+If you installed the Linux host application in a location different from `~/.browserpass`, replace that path in the above script.
+
+3. Change the path in `%localappdata%\browserpass\browserpass-firefox.json` (or `-chrome.json`) to point to `browserpass-wsl.bat`
+
+If your GPG key has a password, the host application running under WSL won't be able to unlock it since it can't interactively prompt for the password. This means you can't get decrypt any passwords unless you've already got the key loaded in gpg-agent.
+As a workaround, you can use the key (`pass website.com`) in a WSL terminal to load the key into gpg-agent. Then browserpass will work until gpg-agent times out.
 
 #### Installing the Chrome extension
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ If you installed the Linux host application in a location different from `~/.bro
 
 3. Change the path in `%localappdata%\browserpass\browserpass-firefox.json` (or `-chrome.json`) to point to `browserpass-wsl.bat`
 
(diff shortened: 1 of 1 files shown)


jtsymon:
Apparently the PR title doesn't link issues 😞 
#192

maximbaz:
Thanks!

1. Extending the `install.ps1` sounds like a good idea, do you want to try implementing it? I won't be able to test this bit personally, but we have a couple of Windows users, I can ping them.
2. Yeah I don't think it's a good idea to implement pin-entry dialog in the extension itself, don't want to touch users' passwords 🙂 But it is a very important part, or people will be confused why nothing is working. Could you maybe mention this in the README as well? Just mention that unfortunately there is such issue, and as a workaround it is required to manually unlock the key, e.g. just show a password inside WSL.

maximbaz:
By the way, how come on WSL both windows and linux host applications are needed? Are they both being used? The `.bat` file only references the linux host app...

jtsymon:
The Windows host application isn't actually needed, but the config files and registry keys are used.
Normally it'd work like
`Browser -> Native Messaging config -> Host Application`
But this does
`Browser -> Native Messaging config (Windows) -> Host Application (Linux)`
So we need the config files from the Windows host application (but not the actual host application), and the Linux host application (but not its config files).

maximbaz:
Huh, makes sense now, thanks!

jtsymon:
Oh I just noticed another problem. My original `.bat` file assumes the Linux host application is in `~/browserpass/`, but that actually depends on where the user put the browserpass files when installing the Linux host application. So the `.bat` file installed with the Windows host application doesn't actually know where to find the Linux host application.
`install.ps1` could actually also install the Linux version under WSL (behind a prompt or flag), but that could get pretty messy (the cleanest option is probably a separate "WSL" zip that contains the Windows config with the Linux host application and a different install script, but that creates extra maintenance work for very few users).
... so maybe we should scrap the "automatically create the `.bat` file" idea, and just say "create a batch file that uses WSL to load the Linux browserpass host application."

maximbaz:
Agreed, let's go with a simple option then. If there will be a public demand for automating then, we will revisit :wink: 

Add the description of the problem with password-protected key and a known workaround, and let's merge it.

jtsymon:
I added the note about password-protected keys, and also added a note about _why_ to use this (hopefully keeps people from accidentally using this when they just want a Windows setup).

jtsymon:
Only problem is now its taking up half the README for an edge case setup 😆

README.md
```diff
@@ -39,6 +39,22 @@ All release files are signed with [this PGP key](https://keybase.io/maximbaz). T
...
+2. Create `%localappdata%\browserpass\browserpass-wsl.bat` with the following contents:
+```
+@echo off
+bash -c ~/.browserpass/browserpass-linux64
+```
+If you installed the Linux host application in a location different from `~/.browserpass`, replace that path in the above script.
+
+3. Change the path in `%localappdata%\browserpass\browserpass-firefox.json` (or `-chrome.json`) to point to `browserpass-wsl.bat`
+
+If your GPG key has a password, the host application running under WSL won't be able to unlock it since it can't interactively prompt for the password. This means you can't get decrypt any passwords unless you've already got the key loaded in gpg-agent.
```
maximbaz: typo `get decrypt` -> `decrypt`

README.md
```diff
@@ -39,6 +39,22 @@ All release files are signed with [this PGP key](https://keybase.io/maximbaz). T
...
+As a workaround, you can use the key (`pass website.com`) in a WSL terminal to load the key into gpg-agent. Then browserpass will work until gpg-agent times out.
```
maximbaz: Maybe add this hint in the end: `(it is possible to configure larger timeouts, check manual for gpg-agent).`

maximbaz:
That is totally fine, and I don't see it as an edge case, I like the idea behind WSL and I hope it gets better / more widespread in the future 😉

maximbaz:
Cool, thanks a lot - merging in a moment.

PR #196, opened 2017-12-19
Match sub directories (fix #182)

### LINKED ISSUE #182: Sub directories not matched
### General information

* Operating system + version: Ubuntu 17.04
* Browser + version: Chrome Version 61.0.3163.100 (Official Build) (64-bit)
* If installed an official release, put a release version: 2.0.8

---

### Exact steps to reproduce the problem

With the following store map:

```
github
├── JoJo
└── Soullivaneuh
    └── pass
```

Search about `github` term on the extension.

### What should happen?

Having `github/JoJo` and `github/Soullivaneuh/pass` result.

### What happened instead?

Just having `github/JoJo`.

apropostech:


pass/disk.go
```diff
@@ -55,7 +58,13 @@ func (s *diskStore) Search(query string) ([]string, error) {
 		return nil, err
 	}
 
+	matches3, err := zglob.GlobFollowSymlinks(s.path + "/**/" + query + "/*/*.gpg")
```
maximbaz: I think you lack `*` after the query, so it should be `"/**/" + query + "*/*/*.gpg"` just like in the first example.

Also, I think since you add this, the first case (`DOMAIN/USERNAME.gpg`) just becomes a special case of this one, so we can remove it.

And finally, please add a unit test (in `pass/disk_test.go`) to show that the code does indeed look into subdirectories now.

apropostech:
Hi,

Thanks for the feedback.

I've added the unit test as you can see, but regarding the query, I'm not getting it to work properly with `"/**/" + query + "*/*/*.gpg"`. It requires an additional `/` to the query for domains without subdirectories. To me the most logical solution seems to be `"/**/" + query + "/**/*.gpg"` matching 0 or more subdirectories but that doesn't seem to work either. I might be missing something though.

When I'm testing it, with the 3 queries, it works properly for all 3 query types.

maximbaz:
Thanks for the test 🙂 I think that the `*` after `query` (i.e. `... " + query + "*/...`) is need so that I can search for "de" and find "debian.org" - looks like we miss such test, could you add it please?

Your most logical solution seems the most logical to me too 🙂 I'll try your branch later today to see if some unit tests fail on my side if I remove the first of the three query types. Which OS are you using, by the way?

apropostech:
Great, thanks.

Maintaining "part-search" would indeed be useful. I've added a `*` to the third test to enable it for subdirectories as well. And a test for part-search on all three queries.

I'm using a mac. Perhaps it's an issue with zglob (on mac) then since we're both seeing the logic in the expression. :)

maximbaz:
I looked at this and came up with the following patch - could you give it a try?

(I don't seem to have access to push to your branch, so download the patch below, unzip it, `cd browserpass` and run `patch -p1 -i ~/Downloads/0001-Use-only-two-matches-fix-tests.patch`)

When you apply the patch, `cd pass` and `go test` - do they show test failures? All tests pass on my Linux, with exactly the expression that you suggested above - if I change it (e.g. try one `*` instead of `**`), then tests will fail.

Let me know!

[patch.zip](https://github.com/dannyvankooten/browserpass/files/1577491/patch.zip)

apropostech:
Thanks for the additional patch. Noticed that I didn't edit the tests to match the last change so fixed that in the latest commit.

I've tried the more logical fix in your patch, and as discussed above. The tests pass but it doesn't work in firefox (haven't tried chrome) so I suppose there is an issue with the string being passed from the browser plug-in. It requires an additional `/` to match.

As it is now, tests pass and it allows for one-level subdirectories. I will continue to try to find the issue that prevents the `*/**/` solution from matching. Let me know if you have any ideas for what it might be.

maximbaz:
I will try the code with my patch in browsers too, to be honest I only tried tests so far 🙂

maximbaz:
@apropostech I pushed one commit to your branch, this adds more tests - I believe this is the expected full behavior for this feature. Could you please review and say if you agree and/or if you have more examples of what should work? 

The tests fail even with `go test` now, but with this I want to establish a goal for us, what should work by the time when we merge, both in tests and in the browsers.

maximbaz:
Browser tests show identical results for me in Chromium and Firefox on Arch Linux.

```
mkdir -p ~/.password-store/abc.org/wiki/work
touch ~/.password-store/abc.org/wiki/work/user
```

Which search terms found the result created above:

- Expected results: `abc`, `wiki`, `work`, `user`.
- Actual results:
  - Latest release:  `work` or `user`.
  - With 3 matches: `wiki`, `work` and `user`.
  - With 2 matches (`*/*/`): `wiki` and `user`.
  - With 2 matches (`*/**/`): `abc` and `user`.

And by the way, it is completely possible that we find a bug in Glob functions, we are using this library and if we have a proof, we can report a bug: https://github.com/mattn/go-zglob

apropostech:
Thanks for the additional tests.

I believe it does indeed confirm the expected behaviour of `"/**/" + query + "*/**/*.gpg"`. When I run the `TestDiskStoreSearchSubDirectories` it fails with 0 results for wiki however which seems different from your results?

Just to confirm, I also tried my 3-test fix again with this test and removed the second-level directory in test_store and it passes that test.

apropostech:
Actually, on second look, I believe we're getting the same results to the tests.

maximbaz:
I looked a bit more, and you know, it does look like a bug in `zglob.GlobFollowSymlinks`, it cannot properly handle our search request. If you come to the same conclusion, could you maybe report a bug (or a PR for them if you feel like it)?

apropostech:
I ran an additional test on go-zglob and it fails when `/**/` needs to match more than one directory. I've reported a bug https://github.com/mattn/go-zglob/issues/15 accordingly.

I also tried a few more variants in the browserpass tests and it seems the `*.gpg` causes problems. `"/**/" + query + "*/**/*` seems to match everything as it should but obviously fails since it also matches the directories.

maximbaz:
Thanks for reporting the issue. If the globbing works as expected without the `*.gpg` suffix, I think we can filter out those records manually after calling the Glob function - would that work?

apropostech:
My error in interpreting the `go test` results. It passes on the first test when searching for abc.org as you stated above but fails for the searches on subdirectories which seems odd. But that means it actually works for matching multiple subdirectories in the first search `"abc.org/wiki/work/user5"` so issue seems to be elsewhere.

Additionally, my data is structured like the first search above and it still doesn't match in browser without an additional `/` at the end of the query or `/*/` if a subdirectory.

Perhaps it's more than one issue that's complicating it.

apropostech:
Since `"/**/" + query + "*/**/*.gpg"` passes the test for matching the domain, perhaps that would be an acceptable implementation? And matching subdirectories in search can be a follow-up if we can confirm that it is dependent on fixing the the zglob library.

Assuming that the issue of not being able to pass the string correctly without an additional `/` from the browser is a separate issue that is.

maximbaz:
I agree that we should merge now and follow-up later, not sure if there is even one person who actually needs deeper match on the subdirectories 🙂

Let's make sure it works on the example provided in #182:

```
github
├── JoJo
└── Soullivaneuh
    └── pass
```

Is this the expected behavior that should work?

- searching for `github` should find `github/JoJo` and `github/Soullivaneuh/pass`
- searching for `JoJo` should find `github/JoJo`
- searching for `Soullivaneuh` should find `github/Soullivaneuh/pass`
- searching for `pass` should find `github/Soullivaneuh/pass`

maximbaz:
3 matches is fine, is there a way to disable the tests that aren't supported yet without removing them? Would like to keep them for future.

apropostech:
I've disabled some of the matches for now and all tests pass. Lets try to go to the bottom of the issue with `/**/` though as it might cause other problems. Should we open a new bug for that?

maximbaz:
Awesome, I'll give it a final round of testing and merge if everything is OK.

Regarding `/**/`, while we only have 1-level subdirectories, is there even an example where things don't work as expected? I couldn't reproduce the issue in the browser that you mention above, or I didn't understand it entirely - if you have an example that doesn't work well with the current code, could you show me the layout of files and directories and your search term?

maximbaz:
Now that the issue with `**` is fixed in the library, let's make this PR solve all cases again? Looks like we can again use two matches only 🙂

To update the dependencies, run `$ dep ensure -update`, this will download the latest version of `zglob` and also update `Gopkg.lock`.

apropostech:
Great! Thanks! I can confirm that all tests pass now with zglob fix and `**` also works properly for me in firefox browser on mac. Have changed branch accordingly so should be good to merge.

Issue #98, opened 2017-06-23
Option to specify password store directory


e-user:
Currently, browserpass aligns with `pass` on the commandline by reading the environment variable `PASSWORD_STORE_DIR` to determine the directory to obtain passwords from.  
However, at least Chromium and Chrome are designed to switch users and contexts within the same instance of the browser, which means these contexts can't use different password stores. On the commandline, the environment variable can simply be set and unset, but for the browser, that's not possible. 

Presumably the most typical use case for this is to have a browser window for work and one for private matters, each using a different Chromium "user". 

Please add a feature to specify the password store directory as an add-on option, so different browser contexts can use different password stores.

rafi:
I also set my `$PASSWORD_STORE_DIR` to somewhere else, had to create a symlink to `~/.password-store`..  It would be awesome if there was an option for this.

maximbaz:
As a workaround, you could try creating different desktop shortcuts for different Chrome user profiles, and then editing them and changing their executable from let's say `chrome` to `PASSWORD_STORE_DIR=/home/user/.password-store chrome` - not sure if it will work, but worth a try.

#77 might also be related depending on how it gets implemented.

zsau:
Any thoughts on this? It would be awesome if we could set different store dirs per [container](https://support.mozilla.org/en-US/kb/containers), though I'm not sure what the official status of containers is right now.

maximbaz:
I'm kind of hoping to see #77 implemented one day in a way that is not really dependent on gopass, but instead just a configurable list in the options window allowing users to define different password store dirs. That way it would solve this request, #77 and #34 in one go.

maximbaz:
Implemented in PR #237, will release after a few days of testing myself. Early feedback is welcome.

Issue #77, opened 2017-05-13
Support gopass mounts


adius:
We use gopass at work and therefore I've mounted the companies password store in my personal one.
This means there is `google.com/john@doe.com` and `company-name/google.com/john@company-name.com`
Those company entries however don't show up in browserpass.

martinhoefling:
What would be the best way to implement this. the browserpass searches for the password store and then in it's subdirectories and does not use pass/gopass for this task. I could imagine to implement the Store interface with methods with shell out to gopass. Any other ideas?

mjrider:
@martinhoefling vendoring the gopass code, and calling the subroutines directly instead of re-implementing the same wheel and/or using shell execution?

maximbaz:
Gopass mounts are not physical subdirectories in `.password-store`, otherwise they would have been discovered by browserpass. On the one hand, I like not reimplementing wheels, on the other hand the current code is not using `pass` since #6, so it is not as simple as just replacing `pass` with `gopass` everywhere in the code. 

Since current code is handling fine one password-store, maybe yet another approach would be to make an option to define paths to multiple password stores somewhere in extension config, and reuse the same code as we have now?

adius:
👉  make an option to define paths to multiple password stores somewhere in extension config

maximbaz:
This would be an awesome next addition to the project. I see this as a list of entries on the options page, with a checkbox next to each entry so that we can configure the list of password stores once, and toggle which ones are needed (such toggling functionality is requested in #98).

The options page was recently created in #95.

The primary challenge here I think is communicating from browser to host app this list of active password stores, and making sure the list doesn't go out of sync between the browser and the host app. Once this list is available in the host app, wrapping the existing code in a loop should not be big of a deal.

So... anyone up for a challenge? 🙂

zsau:
My purposes would be served by a one-store-per-container model. Do others want to have multiple stores active at one time, or maybe the ability to switch between stores on-the-fly without using containers?

maximbaz:
I personally understood this ticket as specifically about many-to-one model, where there are multiple password stores (one personal and one for work), and people want to combine both of them and see credentials from both in the single browser/browserpass instance. 

The gopass is a CLI tool that "virtually" combines many password stores in one, here it is asked about the same thing for browserpass.

erayd:
I concur. I don't actually need this (one password store is fine for my purposes), however if I was using multiple stores, I would want them all active simultaneously.

[...]

Issue #34, opened 2017-01-13
Search passwords only in a specified subfolder of my password-store


mjrider:
i have a shared pass store, sync'd between me and my wife, with folders for my and here private/corporate/financial passwords

it would be awesome if i could have a settings to indicate that all filenames should start with robbert/websites/ 

example:
robbert/websites/github.com and robbert/recovery/github.com are different things, and also have different gpg keys used for encryption

dannyvankooten:
Hey @mjrider,

Are those two separate password stores, both managed via Pass? Or is it one and the same password store, just with different directories?

If it's two different password stores then this should already be possible by making sure your the `PASSWORD_STORE_DIR` environment variable is set properly.

If not then we may be able to add support for this through an environment variable of its own.

mjrider:
@dannyvankooten 1 store(1 git repo) 2 users

maximbaz:
We will have a simple options page soon in https://github.com/dannyvankooten/browserpass/pull/95, this is another place where we could configure such prefix, if not with environment variable.

At the same time I'm keeping an eye on https://github.com/dannyvankooten/browserpass/issues/126, I currently also use prefixes, but I'm wondering if using several password stores with "mounts" feature from gopass is just a better way to do it.

maximbaz:
Implemented in PR #237, will release after a few days of testing myself. Early feedback is welcome.

PR #202, opened 2017-12-30
autosubmit default value switched to false

arslanm:
Users should be able to verify the input before the form's submitted.


maximbaz:
I personally have it also set to `false`, my concern is that this will literally affect everyone who uses the current default of `true` - one day they wake up and find out that autosubmit is not working anymore.

I'm curious, what makes you want to change the default value instead of just changing the option for yourself?

If you really want to proceed with this, I think we should do the change in two stages:
1. Release a version that will write the current value in the local storage (so those people who haven't changed the default of `true` will save `true` in their local storage).
2. Release a new version in a week changing the default, which will now affect only new users.

arslanm:
With local storage disabled, every time you launch a new Firefox instance you need to turn it off on the settings page. This patch is applied to function restore_options() which runs on DOMContentLoaded so it shouldn't affect  current users who have local storage enabled. First time users  and those with no local storage will have autosubmit = false as default.

maximbaz:
What about people who have local storage enabled, but like the default value "true" and thus have never even opened the options page? Their local storage is empty, and changing the default value will affect all of them. 

That's why I was suggesting the two stage process, where we first ensure to save the current value in local storage, and only as a second step change the default that will only affect new users.

maximbaz:
I should also mention that this is not the only piece of code that needs to be changed to adapt the default value, this file is only about the options page. There are also these places, where the default value of `true` actually makes a difference on the behavior of browserpass:

https://github.com/dannyvankooten/browserpass/blob/76805fa2a131c4443c1208ca40b360896bc4dc43/chrome/background.js#L13

https://github.com/dannyvankooten/browserpass/blob/76805fa2a131c4443c1208ca40b360896bc4dc43/chrome/inject.js#L218-L222

arslanm:
I see what you mean. I was under the impression earlier that the auto submit flag was automatically pushed to local storage somewhere else. In any case I think it's practical to have [auto submit flag on the main window](https://scrot.moe/image/6AEyd) rather than on the options page.

maximbaz:
Hey, thanks for looking into this - and happy new year! 🙂

Regarding your last message, I don't really think auto-submit toggle should be moved out to the main window. The main window should contain features that you need to often interact with, and I see the auto-submit toggle as a decision that people make only once, either choose to save time when submitting forms and allow auto-submit, or choose to verify the input and disable auto-submit. Why would you change it more often?

I still think that changing the default value to `false` for new users makes a lot of sense, everyone I personally know who uses browserpass has disabled the auto-submit, including myself. But again, it should be changed for new users only, not affecting the current users.

arslanm:
Happy new year to you too @maximbaz :)

Before migrating to `pass` I was a 1Password user, and before that LastPass. Both apps provides an autosubmit flag that can be changed easily. I don't have LastPass any more but [here's 1Password's menu](https://scrot.moe/image/6Ail0) -- easily accessible.

Users will probably no reason to change Auto Submit behavior as often as one might assume but I think it is better than keeping an option page with a single parameter to change. I'd rather move the parameter to main window and remove the options page unless more parameters are introduced. Regardless, future parameters may be introduced to the main window under a gear icon. But I don't think an extension as simple as this need more params.

Anyway. With this patch I know what the AutoSubmit flag is set to as soon as I click the extension icon and I can change its value with a single click. Yes it'll stay there every time I click the icon but so is the Search input field, or the copy user/pass icons, both of which I rarely use.

arslanm:
With the last commit auto submit flag works correctly. FYI. I think this is all I can contribute to browserpass for now. Happy new year and cheers!

maximbaz:
We have some requests that might add more configuration options, such as #34, #77 and #98, so having an options page is good as it simplifies future improvements. Furthermore, one could argue that copy user/pass does not consume extra space in the main window, and these buttons together with the search box can anyway be used often, much more often than changing auto submit flag. 

So I think having a separate options section for specific things that you change almost never is helpful for us and helps keeping the main window clean. We can discuss the idea to add gear icon which improves accessibility of the options section, but let's not clutter the main screen with rarely used toggle and let's not remove the framework for adding new options. 

Let me know if you are interested in working on bringing the options section to the main window hidden under the gear icon and/or on changing the default value of the toggle, otherwise I'm afraid this PR cannot be accepted as it is.

maximbaz:
I'll take care of changing the default value to `false` for new users myself, [the first stage](https://github.com/dannyvankooten/browserpass/commit/684e4eadc57d5df74d4488775c6fdf22634a1745) will be released soon, and the second stage will be released in a week or two.

So the question remains only about reworking the options section and bringing it into the main window behind the gear icon. If you want to do it, let's discuss how it would look like before writing code, as I'm not entirely convinced that it is worth the effort. Otherwise, let's keep the options as they are today.

maximbaz:
Closing due to no activity.

PR #206, opened 2018-02-06
Correct folders

landswellsong:
Seems like `install.sh` is out of date in relation to folder locations

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -124,10 +124,10 @@ echo "Installing $BROWSER_NAME host config"
 mkdir -p "$TARGET_DIR"
 
 if [ "$BROWSER_NAME" == "Firefox" ]; then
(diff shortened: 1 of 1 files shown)


maximbaz:
Hey @landswellsong, thanks for the PR, but the install.sh is correct 🙂 

If you want to install the official release, grab any of the `browserpass-*` archives, all of them have these files.

If you want to build from sources of the last stable release, grab `browserpass-src.tar.gz` (and not the one that Github created), it has these files too.

If you want to build from the latest master, follow [CONTRIBUTING.md](https://github.com/dannyvankooten/browserpass/blob/master/CONTRIBUTING.md) to do so - specifically the `make js` will generate these files.

landswellsong:
@maximbaz got it thanks

Issue #197, opened 2017-12-19
"Error: Native host has exited" on macOS with Chrome


raxod502:
### General information

* Operating system + version: macOS 10.11.6 El Capitan
* Browser + version: Google Chrome 63.0.3239.84 (Official Build) (64-bit)
* If installed an official release, put a release version: 2.0.8
* Browserpass extension version as reported by your browser: 2.0.8 (browserpass-ce)

---

### Exact steps to reproduce the problem
1. Download and extract `browserpass-darwinx64.zip` from the [v2.0.8 release on GitHub](https://github.com/dannyvankooten/browserpass/releases/tag/2.0.8).
2. Run `install.sh` and select `1) Chrome`.
3. Install [browserpass-ce from Chrome Web Store](https://chrome.google.com/webstore/detail/browserpass-ce/naepdomgkenhinolocfifgehidddafch), version 2.0.8
4. Click on lock icon and select a password (the list is displayed correctly, and search functionality is working).

### What should happen?

I should be prompted for my passphrase via `gpg-agent`.

### What happened instead?

The error message `Error: Native host has exited` is displayed immediately.

maximbaz:
Hey, there are two common reasons on why this can happen on Mac, let's try them first:

1. Do you use GPGTools? https://github.com/dannyvankooten/browserpass/issues/13#issuecomment-332735747
2. Does extension work if you open the browser from terminal instead of using an application launcher?

raxod502:
1. I don't use GPGTools; I use `gnupg@2.0` version 2.0.30 with `gpg-agent` version 2.0.30 installed via Homebrew.
2. Unfortunately, when I try to open Chrome from the terminal (it worked the last time I had to test browserpass, but apparently an update broke it):
  
  ```
  [63763:46339:1220/134456.851401:ERROR:browser_gpu_channel_host_factory.cc(107)] Failed to launch GPU process.
  ```
  I do get a new Chrome window (though not a new entry in the App Switcher), and the problem can be reproduced with no differences in that window.

maximbaz:
OK, do you have `pinentry-mac` installed? Also, what is the output of `$ which gpg` and `$ which gpg2` ?

raxod502:
Yes, I have `pinentry-mac` version 0.9.4 installed via Homebrew.

```
% which gpg
/usr/local/opt/gnupg@2.0/bin/gpg
% which gpg2
/usr/local/opt/gnupg@2.0/bin/gpg2
```

[...]

Issue #32, opened 2017-01-10
add fuzzy searching instead of just globbing


kcolford:

emersion:
That would be a great improvement. Since we already sort-of list all files in the password store, this shouldn't impact performance too much. I'll look if i can do it.

kcolford:
You could add something in so that if you're on `www.facebook.com` then it will match `facebook.com` (and maybe even all subdomains thereof) but keep it in order of most precise.

And then just simple substring matching for the rest would be useful.

emersion:
I think it already does a simple substring matching, doesn't it?

kcolford:
no, it does a prefix matching (typing in `facebook` brings up `facebook.com` but not `www.facebook.com`)

tarciosaraiva:
+1

dannyvankooten:
I like this for subdomains but am not too sure about enabling this for just _any_ prefix (like it does not with suffixes, as a wildcard). For example, I think when searching for "goo" it should return the following.

```
search: goo
results: 
  - google.com
  - goofoo.com
  - accounts.google.com
```

But not "foogoo.com", as I think some kind of naming consistency / rule is important so the search result list doesn't get cluttered. Do you agree?

kcolford:
I definitely agree, although I think we should distinguish between searches
entered by the user and automatic search by domain name. We can't allow a
similar URL being used as phishing, yet we still want to allow the user
some leaway with typos that they enter.

travisghansen:
User searches also appear to be case sensitive.  Can that be changed?

00dani:
When you're on a subdomain like accounts.google.com, passwords for google.com should definitely be found in the automatic search. Currently I have to manually type the site's name for many sites, since my passwords often aren't organised into the exact subdomain the login page lives on.

Perhaps if you're on a domain foo.bar.baz.quux, Browserpass should automatically search for passwords matching "foo.bar.baz.quux", "bar.baz.quux", and "baz.quux", with results that match the longer search strings being ranked higher?

maximbaz:
Are you on the latest browserpass version? I have passwords saved in `personal/google.com/username` and they are being *automatically* discovered on `accounts.google.com`. (If you are using the browserpass installed from Webstore, then you are **not** on the latest version - #117).

*Manual* search capability is not as good and we have two feature requests to improve it:
* #49 is about when you have a password stored as `personal/account.google.com/username`, it is not being discovered by manual search for `google`.
[...]

PR #209, opened 2018-02-08
POC for a couple of alternative search methods.

sophacles:
Both of these use the same overall pattern:
* Set an environment variable with a program for searching
* Shell out from browserpass to run the search program, passing the
query string as a command arg, and returning results in the same format
as built-in search.

NOTE: the shell command is evaluated like:

`sh -c "$VAR {}"`

Where the `{}` is replace with the search string provided from the user in
the browser.

The two methods implemented here are:

1. Custom command to find pass entries and then search them
2. Filter results of an empty string query from the built-in search

Details:

*Custom command* - Set the environment variable
`BROWSERPASS_SEARCH_COMMAND`. This command will be run in a shell, and the
output from stdout will be returned from the DiskStore.Search function -
each line of stdout will be translated to one entry in search results.

An example command:
`BROWSERPASS_SEARCH_COMMAND="find $PASSWORD_STORE_DIR -type f -name *.gpg | sed -e \"s|$PASSWORD_STORE_DIR/\(.*\)\.gpg.*|\1|\" | fzf -f"`

this will walk the password store for files having the `.gpg` extension,
turn them into the name format used by pass/browserpass, and finally
search for the entries using the fzf program.

*Filter* - Set the environment variable `BROWSERPASS_FILTER_COMMAND`. This
command will be run with the search string as above, however the stdin
will be set to a pipe. The results of running the original search
function with an empty-string query will be sent over the pipe to the
fitler program.

An example command:
`BROWSERPASS_FILTER_COMMAND="fzf -f"`

### DIFF
--- a/pass/disk.go
+++ b/pass/disk.go
@@ -1,6 +1,7 @@
 package pass
 
 import (
+	"bufio"
 	"errors"
 	"io"
 	"os"
@@ -8,9 +9,11 @@ import (
 	"sort"
 	"strings"
 
+	"os/exec"
 	"os/user"
 
 	"github.com/mattn/go-zglob"
+	//"github.com/kballard/go-shellquote"
 )
 
 type diskStore struct {
@@ -32,7 +35,6 @@ func defaultStorePath() (string, error) {
 	if err != nil {
 		return "", err
 	}
-
 	path := os.Getenv("PASSWORD_STORE_DIR")
 	if path == "" {
 		path = filepath.Join(usr.HomeDir, ".password-store")
@@ -43,6 +45,112 @@ func defaultStorePath() (string, error) {
 }
 
 func (s *diskStore) Search(query string) ([]string, error) {
+	external_cmd := os.Getenv("BROWSERPASS_SEARCH_COMMAND")
+	filter_cmd := os.Getenv("BROWSERPASS_FILTER_COMMAND")
+
+	var res []string
+	var err error
+
+	if filter_cmd != "" {
+		res, err = s.filterSearch(query, filter_cmd)
+	} else if external_cmd == "" {
+		res, err = s.builtinSearch(query)
+	} else {
+		res, err = s.externalSearch(query, external_cmd)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	result := unique(res)
+	sort.Strings(result)
+
+	return result, nil
+}
+
+func (s *diskStore) filterSearch(query, cmd_str string) ([]string, error) {
+	var result []string
+
+	items, err := s.builtinSearch("")
+	if err != nil {
+		return nil, err
+	}
+	// Commands like fzf and grep will have an exit code of 1, causing an error
+	// include single quotes to avoid shell execution issues
+	cmd_str += " '" + query + "' || true"
+	cmd := exec.Command("/bin/sh", "-c", cmd_str)
+
+	ri, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	rc, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	scanner := bufio.NewScanner(rc)
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	for _, item := range items {
+		str := item + "\n"
+		ri.Write([]byte(str))
+	}
+
+	ri.Close()
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line != "" {
+			result = append(result, line)
+		}
+	}
+
+	if err := cmd.Wait(); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+
+}
+func (s *diskStore) externalSearch(query, cmd_str string) ([]string, error) {
+	var result []string
+
+	// Commands like fzf and grep will have an exit code of 1, causing an error
+	// include single quotes to avoid shell execution issues
+	cmd_str += " '" + query + "' || true"
+	cmd := exec.Command("/bin/sh", "-c", cmd_str)
+
+	rc, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	scanner := bufio.NewScanner(rc)
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line != "" {
+			result = append(result, line)
+		}
+	}
+
+	if err := cmd.Wait(); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func (s *diskStore) builtinSearch(query string) ([]string, error) {
 	// Search:
 	// 	1. DOMAIN/USERNAME.gpg
 	//	2. DOMAIN.gpg
@@ -59,6 +167,7 @@ func (s *diskStore) Search(query string) ([]string, error) {
 	}
 
 	items := append(matches, matches2...)
+
 	for i, path := range items {
 		item, err := filepath.Rel(s.path, path)
 		if err != nil {
@@ -76,10 +185,8 @@ func (s *diskStore) Search(query string) ([]string, error) {
 		items = append(items, newItems...)
 	}
 
-	result := unique(items)
-	sort.Strings(result)
+	return items, nil
 
-	return result, nil
 }
 
 func (s *diskStore) Open(item string) (io.ReadCloser, error) {


sophacles:
In addition to the above I'd like to note a few things:
* This is very much a WIP!
* Im very new to go - this is my first attempt at actually useful code in this language. I'd love to get pointers on anything I'm doing "wrong"
* I personally prefer the "search filter" way of handling things, but I included both as POCs here to see what folks here think of it.

When I run browserpass I tend to set up a wrapper script and have Chrome call that script instead of the binary directly, since my password store and gpg keys are both in non-standard locations, adding additional environment variables is pretty easy in that case :)

Example wrapper script:
```
#!/bin/bash

export PASSWORD_STORE_DIR=/a/path/to/passwords
export GNUPGHOME=/a/path/to/gpg
# Chrome doesn't include /usr/local in the environment it runs
export PATH="$PATH:/usr/local/bin"

#export BROWSERPASS_SEARCH_COMMAND="find $PASSWORD_STORE_DIR -type f -name *.gpg | sed -e \"s|$PASSWORD_STORE_DIR/\(.*\)\.gpg.*|\1|\" | fzf -f"
export BROWSERPASS_FILTER_COMMAND="fzf -f"

$HOME/src/projects/password_management/src/github.com/dannyvankooten/browserpass/browserpass "$@"
exit $?
```

sophacles:
Also note this is relevant for issue #32

maximbaz:
Hey, thanks for the PR 🙂 Before I look into the code, may I ask what use case are you trying to implement with this change? If it's only for #32, I bet you almost nobody will bother creating wrappers and coming up with proper SEARCH_COMMANDs and FILTER_COMMANDs - maybe we better implement #32 natively in the code, so that users just click a toggle in the options or something?

sophacles:
That's a fair point - it is a bit funky. There's a few different things motivating this approach:

* TBH (and to get the negative thing out of the way) I dislike the current glob search enough that I found myself not using browserpass on sites where I have multiple accounts, or sites like banks that tend to redirect to a different sign-in url for credential entry. Instead of complaining, I thought I'd take a pass at solving the problem and this is the way I knew I could get a result that works.  (related: I am not comfortable enough in golang just yet to do anything more complex, this was already pushing my skill level! :) )
* I know a couple people who would prefer regex over fuzzy search (in general not just for browserpass) - in this approach they can have their preferences met by using `grep` as the filter or in the search command instead of `fzf`. 
* If nothing else, this provides a simple way of rapidly iterating several search techniques before committing to a specific native approach.

One more, but it probably deserves a different ticket:

Since I need to set some env vars via the wrapper script anyway, I think browserpass should always be launched through a shell script by default, so i don't have to muck with which binaries the json files point to anyway. Although, as you point out, in-browser configs (or even just a `~/.browserpass` config file) is probably a better way for those folks who aren't unix weirdos like me :)

maximbaz:
Could you please describe the first point in more details? What is your password store structure and what is the behavior of browserpass that you dislike? I use browserpass in all of these cases, and it works really well for me, that's why I'm asking and that's why I want to understand your issue in more detail, maybe I'll have some alternative hints 🙂 

Here's for example how I handle having one account on multiple domains (e.g. `amazon.co.uk` and `amazon.com`, or bank login that redirects somewhere):

1. Create a primary login, e.g:
    ```
    .password-store/
        amazon.com.gpg
    ```
1. Create a symlink for every domain that uses these same credentials, e.g.:
    ```
    .password-store/
        amazon.co.uk.gpg -> amazon.com.gpg
        amazon.com.gpg
    ```

You store credentials in one file, and browserpass follows all symlinks and always finds the proper password.

I actually found myself not using manual search functionality at all, browserpass finds correct credentials 100% of cases.

--------

Environment variables issue: I take it you use macOS? On Linux it is possible to make all of your environment variables available for GUI apps, without using any wrapper scripts. On macOS it's tricky, I know users who don't want wrapper scripts found a way to [patch browser launchers](https://github.com/dannyvankooten/browserpass/issues/197#issuecomment-354586602) so that they source `~/.profile` and load all environment variables on startup.

sophacles:
Oh wow, that link thing is cool and solves several things! It doesn't quite solve all the things though:
( All the points below are really about fuzzy finding vs globbing, not my particular approach to getting fuzzy finding in place) 

* I regularly need to access several different aws accounts via my browser. I have them bookmarked in their url form:
```
$ACCT_ID.signin.aws.amazon.com
```
But that just redirects me to:
```
https://us-east-1.signin.aws.amazon.com/oauth?....
```
That page has an the account number filled in from the redirect. In addition the username I use is different on some of these accounts. (depending on customer details)

So when I open browserpass there's no easy way to differentiate which account to select, so even after doing the symlink thing, I have some N choices. 

What I want to do is type the account number (or part of it) into the search field and have it narrow down my choices to the right one.  Unfortunately builtin search doesn't do that unless I remember to surround it with `*` (like `*8293*` for example). I find this annoying compared to just the numbers.

* A similar case exists for `gmail.com` where I will sign in with personal, work, contract, and organization addresses regularly. I usually store these as:
```
$PASSWORD_STORE_DIR/gmail.com/customername/username
```
 Again, the `*username*` thing is suboptimal. Even worse is when I have multiple entries for that username (across multiple sites) so i need to remember that if i want to limit to gmail, the search string is `*gmail*/**/*username*` or `*gmail*/*cust*`.

* lastly i have a few contracts where the customers aren't fully OAUTH (etc) yet for SSO, but instead just have their apps auth against AD. For these I just store as:

```
$PASSWORD_STORE_DIR/cust/customername/username
```

and when I'm signing into services I just do a search for customername (and again, this is `*customername*` vs `customername`

---------
As for the OSX - I do use it for my laptop, but my desktops are all linux. I saw that zsh patch app thing and am aware that in linux it's not as important because of various .profile/.xinitrc/and so on options, but I try and shoot for an identical config across all my OSes when possible -- I'll probably use the wrapper script in any case, since it means i only need to set up my `~/bin` with darwin vs linux binaries, and not have to worry about additional setup steps on fresh installs.

sophacles:
Awesome thought on the filter/search distinction! It's one of those ideas that once said it's obviously the right thing to do (but before you brought it up I wouldn't have thought of it on my own :D). I'm not sure how this project prefers to do ticket management - should I open a new feature request issue to continue that part of the discussion?

As far as helping with fuzzy finder stuff - I understand your concerns and mostly agree. I'm glad to help implement it in the codebase proper, seems like a good way to keep learning more about golang.

However, as you pointed out above, the wrapper script isn't strictly required to set environment variables for Chrome (etc), thats just how I do it.  In the short term would it make sense to have the `BROWSERPASS_FILTER_COMMAND` environment variable available as a shell-out to do search filtering until in-code fuzzy-find can be completed? (It also leaves the door open for people to use `grep` for regex and for a temporary/hacky solution to #34 as well)

maximbaz:
Created #211 for filtering, added some ideas to that ticket.

The question is now this: do we really need to have the short-term solution _merged_? We can keep the PR open, your code is available to everyone interested, I can even attach some pre-built binaries for all platforms, and we can focus all energy on working on the built-in solution.

Finally, there is also this: https://github.com/dannyvankooten/browserpass/issues/146
And specifically this quote:

> Given full control of the non-native component of the extension, the attacker can extract passwords stored in the configured repository, but can not obtain files elsewhere on the filesystem or reach code execution.

Having `BROWSERPASS_FILTER_COMMAND` in principle allows to execute an arbitrary command. You might say "hey, but this is users' responsibility to make sure their command is secure, properly escaped, etc.", but it is still a violation of that section - until now we offered this protection, guaranteed what exactly the extension can and cannot do, with allowing execution of arbitrary shell command this is becoming false.

sophacles:
Oh neat - I didn't even think through the shell quoting "whack-a-mole" game this PR introduces :( Good call out. (And having read through #146 I applaud the zealous commitment to security of everyones passwords!)

I think if we leave this open, or even just referenced until a built-in fuzzy-find is implemented, that would work well for me and still give people an opportunity to play with it in the interim.

maximbaz:
Let's discuss how to implement fuzzy-find, to agree on some approach before you jump into coding 🙂 

My first idea is that browser extension needs to send an extra parameter to the host app - the search method ("glob" or "fuzzy").

https://github.com/dannyvankooten/browserpass/blob/1e464970a3a3b00a5679482177db7b3d5fe98a95/browserpass.go#L30-L35

The host app should treat the absence of this parameter as "glob", to have some kind of backwards compatibility. 

Browser extension has an options page, now it has only one configuration:

https://github.com/dannyvankooten/browserpass/blob/7db3e8c4fbed348bf80f4d40dc8ae15f3fa56e87/chrome/options.html#L11-L17

We can add a dropdown to select among "glob" and "fuzzy".

As for how to implement actual fuzzy file search... That I don't know 🙂 For glob search we use [go-zglob](https://github.com/mattn/go-zglob).

sophacles:
For the fuzzy search itself, I found a couple libraries that do it:
* https://github.com/renstrom/fuzzysearch
* https://github.com/sahilm/fuzzy/blob/master/fuzzy.go

Both of them (and fuzzy matchers in general) follow general pattern of:

* pass a query and a list of strings to the algorithm
* return a list of matches which is a subset of the list of strings originally passed.
* sometimes provide additional information (e.g. renstrom/fuzzysearch has RankFind, which also includes a levenshtein distance between the match and the result.

Since the whole string is used in the match, It seems like we'll need to do something like this:
1. Walk the `$PASSWORD_STORE_DIR` and generate a list of all candidate `.gpg` files
2. Trim the prefix of `$PASSWORD_STORE_DIR`, and the suffix `.gpg`
3. Apply fuzzy search using the query string and list of trimmed candidates
4. Return the results

Note: The function `filterSearch` in the `pass/disk.go` file in this PR follows these steps. In an earlier experiement I didn't do step 2, which caused some annoyances - specifically since the fuzzy match also was searching on the full `$PASSWORD_STORE_DIR` path, there were many more matches that didn't make a lot of sense. When I applied the cleanup before doing fuzzy search, the results made much more sense.

Further investigation: is there a faster way of walking the directory and getting the cleaned up candidate list than I do in `filterSearch` in this pr. (Specifically the method to beat is:
1. do glob search for `$PASSWORD_STORE_DIR/**/*.gpg`
2a. remove prefixs
2b. remove suffixs
)

maximbaz:
We already have the code to retrieve `.gpg` files and trim prefixes and suffixes 🙂 I think it can be generalized and reused, no need to optimize it for now.

https://github.com/dannyvankooten/browserpass/blob/01c6253c15628e4345bea12df96c568f934eacf3/pass/disk.go#L51-L68

As for the libraries, it would be curious to try them both and see which one returns a more "natural" results... `RankFind` sounds promising.

maximbaz:
Can I close this in favor of #213?

sophacles:
Yep!

PR #212, opened 2018-02-21
Configure install.sh to run non-interactively

eligundry:
I needed Browserpass to be able to install non-interactively so I could automate with my provisioning setup. To do this, I configured the `$BROWSER` variable to be optionally populated via the second argument. In practice, this means you can install the Chrome plugin by running `./install.sh chrome`
 or by running `./install.sh` and selecting `1`.


install.sh
```diff
@@ -83,32 +84,34 @@ if [ -e "$BIN_DIR/browserpass" ]; then
...
-3)
+3|[Ff]irefox)
   BROWSER_NAME="Firefox"
   TARGET_DIR="$TARGET_DIR_FIREFOX"
   ;;
-4)
+4|[Vv]ivaldi)
   BROWSER_NAME="Vivaldi"
   TARGET_DIR="$TARGET_DIR_VIVALDI"
   ;;
```
maximbaz: Could you also update the text in `*)` part? Otherwise this would be a confusing behavior:

```
$ ./install.sh chromium-nightly
Invalid selection. Please select 1-4.
```

eligundry: Done. I appended and force pushed to keep the commit history clean.

maximbaz: Awesome, thanks 🙂

PR #213, opened 2018-02-25
Add builtin fuzzy support to browserpass #32

sophacles:
This adds builtin support for fuzzy finding in the browserpass search
bar. (Issue #32) It also adds options to:

1. Turn on/off the use of fuzzy search from the search bar.
2. Select which fuzzy finding algorithm to use.

The second option allows users to choose between:
* github.com/renstrom/fuzzysearch/fuzzy
* github.com/sahilm/fuzzy

NOTE: the selection of fuzzy algorithm should probably go away before
merging the branch, but it is included right now, since there was a
discussion on #209 about which of those algorithms would be right for
browserpass. This makes switching between the two for evaluation
simpler, until a decision can be made on which to include (or a decision
to keep as-is).

### DIFF
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -90,7 +90,7 @@ function onMessage(request, sender, sendResponse) {
   // object that has current settings. Update this as new settings
   // are added (or old ones removed)
   if (request.action == "getSettings") {
-    const use_fuzzy_search = localStorage.getItem("use_fuzzy_search") == "true";
+    const use_fuzzy_search = localStorage.getItem("use_fuzzy_search") != "false";
     sendResponse({ use_fuzzy_search: use_fuzzy_search})
   }
 }
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -85,6 +85,15 @@ function onMessage(request, sender, sendResponse) {
   if (request.action == "dismissOTP" && sender.tab.id in tabInfos) {
     delete tabInfos[sender.tab.id];
   }
+
+  // allows the local communication to request settings. Returns an
+  // object that has current settings. Update this as new settings
+  // are added (or old ones removed)
+  if (request.action == "getSettings") {
+    const use_fuzzy = localStorage.getItem("use_fuzzy") == "true";
+    const fuzzy_algorithm = localStorage.getItem("fuzzy_algorithm");
+    sendResponse({ use_fuzzy: use_fuzzy, fuzzy_algorithm: fuzzy_algorithm })
+  }
 }
 
 function onTabUpdated(tabId, changeInfo, tab) {
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -90,9 +90,8 @@ function onMessage(request, sender, sendResponse) {
   // object that has current settings. Update this as new settings
   // are added (or old ones removed)
   if (request.action == "getSettings") {
-    const use_fuzzy = localStorage.getItem("use_fuzzy") == "true";
-    const fuzzy_algorithm = localStorage.getItem("fuzzy_algorithm");
-    sendResponse({ use_fuzzy: use_fuzzy, fuzzy_algorithm: fuzzy_algorithm })
+    const use_fuzzy_search = localStorage.getItem("use_fuzzy_search") == "true";
+    sendResponse({ use_fuzzy_search: use_fuzzy_search})
   }
 }
 
--- a/pass/disk.go
+++ b/pass/disk.go
@@ -11,10 +11,15 @@ import (
 	"os/user"
 
 	"github.com/mattn/go-zglob"
+	// Fuzzy matching packages
+	rfuzzy "github.com/renstrom/fuzzysearch/fuzzy"
+	sfuzzy "github.com/sahilm/fuzzy"
 )
 
 type diskStore struct {
-	path string
+	path           string
+	useFuzzy       bool   // Setting to use the fuzzy matcher or the legacy glob matching
+	fuzzyAlgorithm string // Setting to choose the fuzzy algorithm
 }
 
 func NewDefaultStore() (Store, error) {
@@ -23,7 +28,7 @@ func NewDefaultStore() (Store, error) {
 		return nil, err
 	}
 
-	return &diskStore{path}, nil
+	return &diskStore{path, false, ""}, nil
 }
 
 func defaultStorePath() (string, error) {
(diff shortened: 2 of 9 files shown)


sophacles:
I've been running this on my machine - OSX/Chrome for about 1.5 weeks. I find the results of the `sahilm/fuzzy` package to be better as its ranking includes exact sub-matches in it's ranking, whereas the `RankFind` of `renstrom/fuzzy` only uses levenstein distance and is sometime non-intuitive.

Also note before this can be merged:
* [_] A decision on which algorithm to use should be reached
* [_] Testing on Linux and windows should be done with Chrome
* [_] Testing all OSes with firefox should be done.

maximbaz:
This is awesome @sophacles! 

I adore the selector for two fuzzy search libraries, such a cool idea 👍 I tested both, and I definitely agree with you, `sahilm` prioritizes exact sub-matches and that is intuitively what I want. 

I didn't look into the code yet, but one thing you probably should do is add `sort.Strings(result)` in the end of every fuzzy search function in `pass/disk.go`, just like `GlobSearch` has. That way, `sahilm`'s results begin to have on top exactly the same suggestions as `GlobSearch` in exactly the same order, and because of this we could even argue that it makes sense to enable fuzzy search by default - users that search for a certain substring will still see the same results on top, but they will see _more_ results and that way will learn about the new feature of fuzzy search.

I'll continue using this PR for a few days to give it a longer test run, but since both of us agree that `sahilm` library is the way to go, feel free to remove it from PR now.

maximbaz:
This is awesome @sophacles! 

I adore the selector for two fuzzy search libraries, such a cool idea 👍 I tested both, and I definitely agree with you, `sahilm` prioritizes exact sub-matches and that is intuitively what I want. 

~~I didn't look into the code yet, but one thing you probably should do is add `sort.Strings(result)` in the end of every fuzzy search function in `pass/disk.go`, just like `GlobSearch` has. That way, `sahilm`'s results begin to have on top exactly the same suggestions as `GlobSearch` in exactly the same order, and because of this we could even argue that it makes sense to enable fuzzy search by default - users that search for a certain substring will still see the same results on top, but they will see _more_ results and that way will learn about the new feature of fuzzy search.~~ 
No, that is completely false, this ignores the order computed by fuzzy search and is not what we want. 

But then it's a little disappointing that those results that have exact sub-matches are not sorted, and people will notice the difference and be confused if we enable fuzzy search by default. 

Exact sub-matches are important because when you open a URL and click on browserpass icon, the domain name will be searched (now with fuzzy search) and most likely only the credentials from this domain will be discovered, and it would be really nice if such "automatic domain name search" provided the same results in the same order in Glob and Fuzzy modes (if we want to enable fuzzy search by default).

For example, when I open `github.com` and click on browserpass, with Glob I had: 

```
personal/github.com
work/github.com
```

With `sahilm` fuzzy search I have now:

```
work/github.com
personal/github.com
```

My muscle memory is already used to hit <kbd>Tab</kbd><kbd>Enter</kbd> to login with personal github, now I will suddenly login with work account... 🤔 

------------

I'll continue using this PR for a few days to give it a longer test run, but since both of us agree that `sahilm` library is the way to go, feel free to remove the other one from PR now.

maximbaz:
Looking at the code of `sahilm/fuzzy` I see that this is their fundamental property, matches earlier in the string win. In my example, `github.com` in string `work/github.com` begins to match since index `4`, while in string `personal/github.com` it begins to match only since index `8`, therefore `work/github.com` ranks first.

maximbaz:
I found one issue:

Create `domain.com.gpg` in your `.password-store`, go to `sub.domain.com` in the browser, click on browserpass icon.

With glob search, `domain.com` will be discovered, with fuzzy search it is not discovered.

I think there is some code in GlobSearch that unwraps the query string (in this case `sub.domain.com`) into parent domains, perhaps it's not happening with fuzzy search.

sophacles:
@maximbaz Good catch on that subdomain issue!

I think its all the way fixed now. I solved it by adding a new action called `match_domain` in the communication between the plugin and the native client. The handling for this will use glob search always, while normal user entered search will use whichever method is in the settings.

I also removed the renstrom algorithm, and the settings for fuzzy algorithm selection.

maximbaz:
This sounds like a reasonable approach 👍 I'll keep testing and will also review code later today.

pass/disk.go
```diff
@@ -8,13 +8,18 @@ import (
...
 
 	"github.com/mattn/go-zglob"
+	sfuzzy "github.com/sahilm/fuzzy"
 )
 
 type diskStore struct {
-	path string
+	path           string
+	useFuzzy       bool   // Setting to use the fuzzy matcher or the legacy glob matching
+	fuzzyAlgorithm string // Setting to choose the fuzzy algorithm
```
maximbaz: obsolete

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
...
+		//todo validate path exists
+		s.path = *path
+	}
+	if use_fuzzy != nil {
+		s.useFuzzy = *use_fuzzy
+	}
+	return nil
+}
+
+// Do a search. Will call into the correct algoritm (glob, fuzzy (renstrom) or fuzzy (sahilm)
```
maximbaz: Will use glob or fuzzy search based on the settings ...

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
...
+// based on the settings present in the diskStore struct
 func (s *diskStore) Search(query string) ([]string, error) {
+	// default legacy glob search
```
maximbaz: let's not use "legacy" word :)

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
...
+	if !s.useFuzzy {
+		fmt.Fprintf(os.Stderr, "Doing globsearch\n")
+		return s.GlobSearch(query)
+	} else {
+		fmt.Fprintf(os.Stderr, "Doing fuzzy\n")
```
maximbaz: let's remove `fmt.Fprintf` calls

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
...
+		return s.sahilmFuzzySerach(query)
+	}
+}
+
+// Fuzzy searches first get a list of all pass entries by doing a glob search
+// for the empty string, then apply appropriate logic to convert results to
+// a slice of strings, finally returning all of the unique entries.
+func (s *diskStore) sahilmFuzzySerach(query string) ([]string, error) {
```
maximbaz: rename to FuzzySearch

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
...
+
+	// The resulting match struct does not copy the strings, but rather
+	// provides the index to the original array. Copy those strings
+	// into the result slice
+	matches := sfuzzy.Find(query, file_list)
+	for _, match := range matches {
+		items = append(items, file_list[match.Index])
+	}
+
+	result := unique(items)
```
maximbaz: GlobSearch already took care of duplicates, I don't think there could be duplicates anymore, so should be safe to remove

pass/disk.go
```diff
@@ -42,7 +47,55 @@ func defaultStorePath() (string, error) {
 	return filepath.EvalSymlinks(path)
 }
 
+// Set the configuration options for password matching.
+func (s *diskStore) SetConfig(path *string, use_fuzzy *bool) error {
```
maximbaz: I don't see any place where `path` is not nil, is this an obsolete parameter?

sophacles: Oh thanks for calling this out, I forgot to comment on it!

I have that in place because it feels like the Right Thing to have the path to password store be changeable via the `SetConfig` method. I went back and forth when writing it because there's two ways to view the semantics of `SetConfig`:
* `SetConfig` updates any settings that affect the password store wrapper (therefore keep `path`)
* `SetConfig` applies any app settings (therefore remove `path` for now - but i'm thinking of tackling #98 next which will re-introduce this parameter if removed)

That said I have no strong preference on it in terms of this PR.

sophacles:
@maximbaz - thanks for the review. I have pushed up some changes addressing your comments so far.

pass/disk.go
maximbaz: Let's leave it out for this PR and introduce in a separate one, just for clarity in git history.

#98 is a good issue to tackle next, actually there was #77 (which now opens 404 page for me? 😕 ), if I remember correctly it suggested to allow configuring multiple password store directories (e.g. one for work and one personal), and even toggle the list in options (i.e. temporarily disable work password store without removing it from options page). Something like this:

---------------

Password store directories:

- [x] `~/.password-store`
- [x] `~/.password-store-work` 

<kbd>Add</kbd>

--------------

 Anyhow, this we can discuss separately, I just wanted to let you know 🙂

chrome/background.js
```diff
@@ -85,6 +85,14 @@ function onMessage(request, sender, sendResponse) {
   if (request.action == "dismissOTP" && sender.tab.id in tabInfos) {
     delete tabInfos[sender.tab.id];
   }
+
+  // allows the local communication to request settings. Returns an
+  // object that has current settings. Update this as new settings
+  // are added (or old ones removed)
+  if (request.action == "getSettings") {
+    const use_fuzzy_search = localStorage.getItem("use_fuzzy_search") == "true";
```
maximbaz: Since I see you agreed to make fuzzy search the default, change here too: `!= "false"`

pass/disk.go
sophacles: OK sounds good. FYI I'm also getting a 404 for #77 oddly. 

I'll move discussion of password dir settings to #98 when I dive into that.

sophacles:
@maximbaz re testing: I can whip up some tests for the golang part tonight (and maybe over the next couple days depending on how much learning I need to do :) ).

For the plugin side though - I have no idea how to do tests for a plugin, and don't see any existing already to build from... Any pointers?

maximbaz:
A friendly ping @sophacles 🙂

sophacles:
@maximbaz - sorry about that! Got tied up with $DAYJOB things and this slipped my mind.

Just checked in some tests for the pass module! Not sure how to go about doing testing on the changes I've made to `browserpass.go` - input?

maximbaz:
Thanks a lot! I think this is enough, we will live with manual tests for `browserpass.go` for now, I'm mainly concerned with keeping regressions away from searching and matching functionality.

Also, I have been using this branch on a daily basis since you created this PR, and I did not have any issues. 

I'm going to merge this now and release a new versions - I'm keen to get users' feedback since fuzzy search will be enabled by default 😉

pass/disk.go
maximbaz: @sophacles just curious, do you think you will have time to work on #98? No pressure! 😄

Issue #211, opened 2018-02-09
Filter the list of already discovered passwords


maximbaz:
I open `github.com`, click on browserpass and see two results: `github.com/work` and `github.com/personal`. I'd like to filter this list down, type `wo` and leave only `github.com/work` in place. However by doing this I don't want to add `wordpress.com` to the list, I'm not _searching_, I'm _filtering_ passwords here.

These are the initial requirements that I came up with, feel free to comment/suggest:
- The filtering should be fuzzy and only fuzzy, so I can type `wk` and it will also find `github.com/work`. It doesn't make sense to support grep or glob, the list being filtered is likely small, so there is no concern to "find too many results".
- The filtering should be done on javascript side, no need to make calls to backed, we already have all info we need.
- I think the same search bar can be reused for search and filtering, I'm thinking about some having a toggle for modes. The default behavior can be like that:
    - If browserpass is opened and there are no matches, the mode is "search".
    - If browserpass is opened and there are results, the mode is "filter".
    - There needs to be some kind of button/icon next to the search bar to toggle the mode.
    - Placeholder can reflect the current mode, now it says "Search password...", can also say "Filter passwords...".
- The filtering should be in real time, I type `w` and the list is immediately being filtered, no pressing "Enter". If there is only one result left, pressing "Enter" should select it (this will also fix #40).

sophacles:
The filter/search differentiation and cancelling filter to return to search can be handled like this:
 (reusing your example above)
* When there is a domain match, the domain can be displayed in the search box, with the url text highlighted having a background color. (and maybe other fancy decorations :) ).
* when the user types in the search box, it fuzzy filters in the domain results.
* If there is no user-entered text in the search box (so just the highlighted domain name), and the user presses backspace, search mode reverts from filter to full search, and text put into the search bar runs through full search.

sophacles:
Regarding:

> The filtering should be in real time, I type w and the list is immediately being filtered, no pressing "Enter". If there is only one result left, pressing "Enter" should select it (this will also fix #40).

Can the keypress handler also be used for normal search to do real-time search as well (based on filter/search mode setting)?  I know that's a different ticket probably, but bringing it up here as it may influence how the code for filter is designed.

maximbaz:
I like the idea of how you describe filter/search, it reminds me of how Chromium draws search bar for custom searches:

[...]

PR #220, opened 2018-03-19
Implement as-you-type filtering

### LINKED ISSUE #40: Enter key to autocomplete

erayd:
If there are search results present:
 * Typing in the search field will filter the results.
 * Pressing <kbd>enter</kbd> will autofill the first entry.
 * Pressing <kbd>enter</kbd> when there are no matching entries will run a normal search.

If there are search results present, no filter has been entered, and <kbd>Backspace</kbd> is pressed:
 * All search results are cleared, as is the search input box.
 * Pressing <kbd>enter</kbd> will run a normal search.

The reason for the <kbd>Backspace</kbd> feature is to allow the user to explicitly run a manual search instead of filtering, even when the automatic domain search returns matches.

If the search results are the outcome of a manual search, and no filter text has been entered, then pressing <kbd>enter</kbd> will **not** auto-fill the first entry, but will instead re-run the search. This is to prevent the user accidentally filling a login form by pressing <kbd>enter</kbd> twice when running a manual search.

This PR closes #40 and #211.

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -97,11 +97,6 @@ function view() {
 }
 
 function filterLogins(e) {
-  // remove executed search from input field
-  if (!fillOnSubmit && e.target.value.indexOf(domain) === 0) {
-    e.target.value = e.target.value.substr(domain.length);
-  }
-
   // use fuzzy search to filter results
   var filter = e.target.value.trim().split(/[\s\/]+/);
   if (filter.length > 0) {
@@ -132,10 +127,10 @@ function filterLogins(e) {
 
 function searchKeyHandler(e) {
   // switch to search mode if '\' is pressed and no filter text has been entered
-  if (e.code == "Backspace" && logins && logins.length > 0 && (!e.target.value.length || e.target.value == domain)) {
+  if (e.code == "Backspace" && logins && logins.length > 0 && e.target.value.length == 0) {
     e.preventDefault();
     logins = resultLogins = [];
-    e.target.value = '';
+    e.target.value = fillOnSubmit ? '' : domain;
     showFilterHint(false);
   }
 }
@@ -207,7 +202,10 @@ function searchPassword(_domain, action="search", useFillOnSubmit=true) {
           logins = resultLogins = response ? response : [];
           document.getElementById("filter-search").textContent = domain;
           fillOnSubmit = useFillOnSubmit && logins && logins.length > 0;
-          showFilterHint(fillOnSubmit);
+          if (logins && logins.length > 0) {
+            showFilterHint(true);
+            document.getElementById("search-field").value = '';
+          }
           m.redraw();
         }
       );
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var m = require("mithril");
+var Fuse = require("fuse.js");
 var app = "com.dannyvankooten.browserpass";
 var activeTab;
 var searching = false;
@@ -101,21 +102,25 @@ function filterLogins(e) {
     e.target.value = e.target.value.substr(domain.length);
   }
 
-  // get filter array and reset available logins
-  var filter = e.target.value.split(/[\s\/]+/);
-  logins = resultLogins.slice(0);
-
+  // use fuse.js fuzzy search to filter results
+  var filter = e.target.value.trim();
   if (filter.length > 0) {
-    // filter logins by each filter element
-    logins:
-    for (var i = logins.length - 1; i >= 0; i--) {
-      for (var j in filter) {
-        if (!~logins[i].indexOf(filter[j])) {
-          logins.splice(i, 1);
-          continue logins;
-        }
-      }
-    }
+    logins = [];
+    var fuseOptions = {
+      shouldSort: true,
+      tokenize: true,
+      matchAllTokens: true,
+      threshold: 0.4,
+      location: 0,
+      distance: 100,
+      maxPatternLength: 32,
+      minMatchCharLength: 1,
+      keys: undefined
+    };
+    var fuse = new Fuse(resultLogins, fuseOptions);
+    fuse.search(filter).forEach(function(i) {
+      logins.push(resultLogins[i]);
+    });
 
     // fill login forms on submit rather than initiating a search
     fillOnSubmit = logins.length > 0;
@@ -125,7 +130,7 @@ function filterLogins(e) {
   m.redraw();
 
   // show / hide the filter hint
-  showFilterHint(filter.length > 0 && logins.length);
+  showFilterHint(e.target.value.length > 0 && logins.length);
 }
 
 function searchKeyHandler(e) {
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -76,7 +76,7 @@ function view() {
               type: "text",
               id: "search-field",
               name: "s",
-              placeholder: "Search password..",
+              placeholder: "Search passwords..",
               autocomplete: "off",
               autofocus: "on",
               oninput: filterLogins
@@ -147,10 +147,13 @@ function searchKeyHandler(e) {
 
 function showFilterHint(show=true) {
   var filterHint = document.getElementById("filter-search");
+  var searchField = document.getElementById("search-field");
   if (show) {
     filterHint.style.display = "block";
+    searchField.setAttribute("placeholder", "Refine search...");
   } else {
     filterHint.style.display = "none";
+    searchField.setAttribute("placeholder", "Search passwords...");
   }
 }
 
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -126,8 +126,8 @@ function filterLogins(e) {
 }
 
 function searchKeyHandler(e) {
-  // switch to search mode if '\' is pressed and no filter text has been entered
-  if (e.code == "Backspace" && logins && logins.length > 0 && e.target.value.length == 0) {
+  // switch to search mode if backspace is pressed and no filter text has been entered
+  if (e.code == "Backspace" && logins.length > 0 && e.target.value.length == 0) {
     e.preventDefault();
     logins = resultLogins = [];
     e.target.value = fillOnSubmit ? '' : domain;
@@ -150,7 +150,7 @@ function showFilterHint(show=true) {
 function submitSearchForm(e) {
   e.preventDefault();
 
-  if (fillOnSubmit && logins && logins.length > 0) {
+  if (fillOnSubmit && logins.length > 0) {
     // fill using the first result
     getLoginData.bind(logins[0])();
   } else {
@@ -201,7 +201,7 @@ function searchPassword(_domain, action="search", useFillOnSubmit=true) {
           searching = false;
           logins = resultLogins = response ? response : [];
           document.getElementById("filter-search").textContent = domain;
-          fillOnSubmit = useFillOnSubmit && logins && logins.length > 0;
+          fillOnSubmit = useFillOnSubmit && logins.length > 0;
           if (logins && logins.length > 0) {
             showFilterHint(true);
             document.getElementById("search-field").value = '';
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -103,11 +103,17 @@ function filterLogins(e) {
   }
 
   // use fuzzy search to filter results
-  var filter = e.target.value.trim();
+  var filter = e.target.value.trim().split(/[\s\/]+/);
   if (filter.length > 0) {
-    logins = [];
-    FuzzySort.go(filter, resultLogins, {allowTypo: false}).forEach(function(result) {
-      logins.push(result.target);
+    logins = resultLogins.slice(0);
+    filter.forEach(function(word) {
+      if (word.length > 0) {
+        var refine = [];
+        FuzzySort.go(word, logins, {allowTypo: false}).forEach(function(result) {
+          refine.push(result.target);
+        });
+        logins = refine.slice(0);
+      }
     });
 
     // fill login forms on submit rather than initiating a search
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -295,14 +295,16 @@ function keyHandler(e) {
       switchFocus("div.entry:first-child > .login", "nextElementSibling");
       break;
     case "c":
-      if (e.ctrlKey) {
+      if (e.target.id != "search-field" && e.ctrlKey) {
         document.activeElement["nextElementSibling"][
           "nextElementSibling"
         ].click();
       }
       break;
     case "C":
-      document.activeElement["nextElementSibling"].click();
+      if (e.target.id != "search-field") {
+        document.activeElement["nextElementSibling"].click();
+      }
       break;
   }
 }
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -124,6 +124,8 @@ function filterLogins(e) {
 
     // fill login forms on submit rather than initiating a search
     fillOnSubmit = logins.length > 0;
+  } else {
+    logins = resultLogins.slice(0);
   }
 
   // redraw the list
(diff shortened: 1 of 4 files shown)


maximbaz:
Awesome stuff! My initial thoughts:

- I'd like to have a visual indication of which mode we are in, filtering or searching. An interesting idea was suggested in #211, show an indicator on the left side inside of the search box (and clear it with <kbd>Backspace</kbd> instead of <kbd>\\</kbd>, because that's how Chromium behaves). Doesn't need to be fancy looking, but should be "not ugly" 🙂
- Filtering should be fuzzy - again, no need for fancy external libraries, don't change the order of search results, but something very basic, like `wk` should find `work`.

erayd:
@maximbaz 
> I'd like to have a visual indication of which mode we are in, filtering or searching.

No problem - I'll add this.

> clear it with <kbd>Backspace</kbd> instead of <kbd>\\</kbd>

Chrome doesn't appear to fire the `onkeypress` event for backspace in the search field, although it does for any other key. Do you know how to catch this?

> Filtering should be fuzzy - again, no need for fancy external libraries, don't change the order of search results, but something very basic, like wk should find work.

I strongly dislike this idea. The outcome of fuzzy searches tend to be non-obvious, and when I'm wanting to quickly filter a list, I need it to be predictable.

The way it works now is substring matching for each word in the filter, separated by either whitespace or <kbd>/</kbd>. Is there a reason this is undesirable? I'm struggling to understand how fuzzy searching is desirable in a quick-filter.

erayd:
> Chrome doesn't appear to fire the onkeypress event for backspace in the search field, although it does for any other key. Do you know how to catch this?

Never mind - `onkeydown` **does** catch it, so I'll just use that.

maximbaz:
My experience is a lot different with fuzzy search, I find it very useful - also, when done properly, typing "work" should find the "work" entry as the top one, so you would not experience any difference comparing to the substring implementation. Try searching for full substrings in the domain search in v2.0.12, it is fuzzy by default, but nevertheless searching for substrings should put the most expected results on top. However when you get used to typing "wk", you will be pleased that it works too 🙂

However I realize that perhaps a good fuzzy search is not easy to implement and probably requires some dependency to be working acceptably well. So, a substring search is definitely preferred over a bad-working fuzzy search, but a good-working fuzzy search is preferred over the substring search 🙂

And if you are still not convinced, I'm fine with implementing a configuration option, whether filtering should be fuzzy or not. Doesn't even have to be part of this PR.

erayd:
> However I realize that perhaps a good fuzzy search is not easy to implement and probably requires some dependency to be working acceptably well. So, a substring search is definitely preferred over a bad-working fuzzy search, but a good-working fuzzy search is preferred over the substring search.

Do you have a library to recommend for this, that you would be happy to have included in the extension?

> And if you are still not convinced, I'm fine with implementing a configuration option, whether filtering should be fuzzy or not. Doesn't even have to be part of this PR.

If you're happy with a config option, then I have no problem with the filtering being fuzzy, even by default. Just so long as I can turn the fuzzy stuff off :-).

maximbaz:
I'm using this PR and will be publishing new findings as I find something:

- [ ] Open browserpass, it is filtering mode, type something so that no results are left - the indicator of "filter mode" is gone, but should stay.
- [ ] Open browserpass, it is in filtering mode, press <kbd>Backspace</kbd> to switch to search mode, type "githud", press <kbd>Enter</kbd>, press <kbd>Backspace</kbd> once (in attempt to fix the typo and change "githud" to "github") - the entire search field is cleared, but only the last char should be removed.

erayd:
> Open browserpass, it is filtering mode, type something so that no results are left - the indicator of "filter mode" is gone, but should stay.

This is deliberate - if you filter out all results (i.e. to make the result list empty), it switches to regular search mode. That's why the hint vanishes. Can you explain what the desired behavior should be instead?

Typo fixing - good catch. Need to think about how to resolve that one.

I've swiched to fuzzy search using fuse.js - does how I've implemented it fit what you were expecting?

erayd:
OK, I've sorted the typo thing. Is that now working as you expected it to work?

maximbaz:
> This is deliberate - if you filter out all results (i.e. to make the result list empty), it switches to regular search mode. That's why the hint vanishes. Can you explain what the desired behavior should be instead?

Aaaah, I see - no, keep it like this, very useful indeed. I just didn't realize that it switches to the search mode, I thought it is still in the filter mode 🙂

I'll test the fuzzy search later today 👍

maximbaz:
Regarding making fuzzy filter configurable, there is already an option for fuzzy search, I'm wondering if it makes sense to reuse that same option for fuzzy filtering, or introduce a new option. Is it possible that a user would like to disable fuzzy search, but enable fuzzy filtering, or the other way around? Consider that it is also not obvious for people unfamiliar with the code to understand what's the difference between option "Fuzzy search" and "Fuzzy filter". What do you recommend?

erayd:
I found that making the filter a little more strict was enough that I was happy with fuzzy filtering as the sole option. So I would be happy to leave it as-is, and not worry about substring matching.

If you think that it would be better to include both filter methods, then yes - using the existing option for fuzzy search seems like a good idea.

maximbaz:
OK, let's keep the fuzzy filtering as the sole option for now.

I'm only unhappy with the library, I saw people complain about it, and now I see why:

fuzzy

`ig` should already find no result, and no additions to `ig` should add results.

I'm reviewing a few options now, what do you think about this library?

https://github.com/farzher/fuzzysort

If we choose it, let's disallow typos to make it strict (i.e. `allowTypo: false` if it's true by default).

I haven't tested it myself yet in much details, but the demo seems quite good. Also this library is maintained now (contrary to some other existing libraries), which I like.

erayd:
I see what you mean - the behavior in your example is horrible! I'll take a look at the one you're suggesting now, hang on.

erayd:
That library seems to be overly strict with filtering on patterns that aren't at the start of a word. For example, `hub` does not match `github`. This is problematic.

maximbaz:
Hmm this is not happening on their demo page at least...

erayd:
It happens as soon as you set a strictness threshold. Looks like it might be OK with no threshold set though - have just commited it so you can have a look. What do you think?

erayd:
I might need to do an iterated version; seems it can't handle multi-word searches.

maximbaz:
It works much better, but let's both test a bit more. Good finding on multi-word searches, but here's a workaround that you might like: remove all whitespaces from the user input 🙂 It seems library doesn't like "work github", but handles pretty well the filter for "workgithub".

erayd:
Do you mind if we keep it as a multi-word loop (i.e. how I have it now)? Because just stripping the whitespace can't handle out-of-order words, whereas refining in a loop can.

maximbaz:
I don't mind at all, if you like it, let's keep it 👍 

From the usage point of view, this looks pretty solid. I'll review the code a bit later, and will test more in meantime.

Found this on chrome://extensions, could you please have a look?

image

erayd:
> Found this on chrome://extensions, could you please have a look?

Could you please clarify how / where to reproduce that, if it's still an issue? I've just pushed a commit that fixes the type, but I would like to test to ensure that solved the issue.

maximbaz:
I didn't see when exactly that happened, but I think you most probably did fix it with the last commit. I cleared all the errors and will be testing the code in its latest stage, if any errors appear I'll compose the repro steps.

erayd:
Sounds good - thanks :-).

chrome/script.browserify.js
```diff
@@ -86,15 +96,77 @@ function view() {
...
+  // redraw the list
+  m.redraw();
+
+  // show / hide the filter hint
+  showFilterHint(logins && logins.length);
+}
+
+function searchKeyHandler(e) {
+  // switch to search mode if '\' is pressed and no filter text has been entered
+  if (e.code == "Backspace" && logins && logins.length > 0 && (!e.target.value.length || e.target.value == domain)) {
```
maximbaz: - [ ] I think `&& logins` is not needed anymore since it can never be null
- [ ] `switch to search mode if '\' is pressed` - needs to be `Backspace`

chrome/script.browserify.js
```diff
@@ -86,15 +96,77 @@ function view() {
...
+}
+
 function submitSearchForm(e) {
   e.preventDefault();
 
-  // don't search without input.
-  if (!this.s.value.length) {
-    return;
-  }
+  if (fillOnSubmit && logins && logins.length > 0) {
```
maximbaz: I think `&& logins` is not needed anymore since it can never be null

chrome/script.browserify.js
```diff
@@ -132,7 +204,10 @@ function searchPassword(_domain, action="search") {
           }
 
           searching = false;
-          logins = response;
+          logins = resultLogins = response ? response : [];
+          document.getElementById("filter-search").textContent = domain;
+          fillOnSubmit = useFillOnSubmit && logins && logins.length > 0;
```
maximbaz: `&& logins`

chrome/script.browserify.js
```diff
@@ -222,14 +298,16 @@ function keyHandler(e) {
       switchFocus("div.entry:first-child > .login", "nextElementSibling");
       break;
     case "c":
-      if (e.ctrlKey) {
+      if (e.target.id != "search-field" && e.ctrlKey) {
```
maximbaz: both `copy` shortcuts seem to not work for me anymore, do they work for you?

chrome/script.browserify.js
```diff
@@ -86,15 +96,77 @@ function view() {
   ]);
 }
 
+function filterLogins(e) {
+  // remove executed search from input field
+  if (!fillOnSubmit && e.target.value.indexOf(domain) === 0) {
+    e.target.value = e.target.value.substr(domain.length);
```
maximbaz: could you please explain to me the purpose of this piece?

It works weirdly...:
1. Open browserpass, it is in filter mode
2. Press Backspace to switch to search mode
3. Type "gith", press Enter
4. Press Backspace, everything is cleared (difficult to fix typo - not because of this code block, but still a valid issue)
5. Slowly type "gith" again, when you enter "h" of "gith" the field gets cleared again (because of this code block)

chrome/script.browserify.js
```diff
@@ -86,15 +96,77 @@ function view() {
...
+
+  // show / hide the filter hint
+  showFilterHint(logins && logins.length);
+}
+
+function searchKeyHandler(e) {
+  // switch to search mode if '\' is pressed and no filter text has been entered
+  if (e.code == "Backspace" && logins && logins.length > 0 && (!e.target.value.length || e.target.value == domain)) {
+    e.preventDefault();
+    logins = resultLogins = [];
```
maximbaz: This is very minor thing and most probably we shouldn't do anything here, but posting in case you have some idea how to avoid the confusion. If not, don't worry about this.

I press Backspace and I am told that there are no passwords for github.com.

backspace

chrome/script.browserify.js
erayd: It's a bug. Was half of a previous solution, which I have now removed.

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
erayd: Investigating...

chrome/script.browserify.js
erayd: I will fix this.

chrome/script.browserify.js
```diff
@@ -86,15 +96,72 @@ function view() {
...
+  } else {
+    // reset the result list if the filter is empty
+    logins = resultLogins.slice(0);
+  }
+
+  // redraw the list
+  m.redraw();
+
+  // show / hide the filter hint
+  showFilterHint(logins && logins.length);
```
maximbaz: one more place with `logins &&`

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
erayd: These seem to be working for me. Perhaps fixed by one of the other commits?

Are you still having issues, or are they working for you now?

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
maximbaz: Emm they actually don't even work for me in release version... okay, nevermind, I'll deal with them separately.

erayd: One thought - is the item actually selected when you use the copy shortcuts? I did change the behavior so that they did not run when the search box was selected, because they conflicted with entering filter text (i.e. typing a capital C resulted in a copy-username-and-close behavior, which is obviously undesirable). That's what the `target-id` check is for.

chrome/script.browserify.js
```diff
@@ -132,7 +200,13 @@ function searchPassword(_domain, action="search") {
           }
 
           searching = false;
-          logins = response;
+          logins = resultLogins = response ? response : [];
+          document.getElementById("filter-search").textContent = domain;
+          fillOnSubmit = useFillOnSubmit && logins.length > 0;
+          if (logins && logins.length > 0) {
```
maximbaz: one more place with `logins &&`... sorry 😃

erayd: Done.

maximbaz:
Is it intentional that now after every search the form is switched back in the filter mode? I didn't notice in which commit this appeared, but I actually think it is interesting 🙂

auto-filter

erayd:
> Is it intentional that now after every search the form is switched back in the filter mode?

Yes. It will switch to filter mode if the search returns one or more results, but will stay in search mode if no results are found.

If you hit backspace immediately after it switches to filter mode from a manual search, it will go back to editing the search query.

erayd:
By the way, would you like me to squash this PR into a single commit? Or leave it as-is?

maximbaz:
Very cool. Seems pretty solid to me, is there anything else you want to do or that is not done yet?

I just tested in Firefox, there are no noticeable issues as well.

Leave as it is, Github will squash when merging, but in PR we will have history of commits.

erayd:
> Very cool. Seems pretty solid to me, is there anything else you want to do or that is not done yet?

Nope, I'm pretty happy with it now. It covers all my use-cases, and I can't currently think of anything else it ought to do that others might want from it.

> Leave as it is

Will do :-).

erayd:
Is there anything else you'd like me to do to this PR before it's ready to merge?

erayd:
Thank you too - your input resulted in it being a lot better than it was when I initially opened the PR!

jangari:
Great work guys!

Question on usage: filter-as-you-type only seems to work on the search results, not where you have no matching domain, and so no automatically populated search results. Is that right?

When I saw this discussion yesterday I assumed it would add the ability to type to filter all password-store files when no search was active. So, the following _should_ (under my assumptions) show me all passwords matching 'wor', i.e. everything in my Work folder, plus whatever else (fuzzily) matches that pattern:
image

Am I off the mark here?

sophacles:
@maximbaz @erayd I just got around to pulling these changes and playing with them -- this is AWESOME! Well done :)

erayd:
@jangari 
That's how it's supposed to work. Once you search, it will then start filtering results.

If you want to filter your entire database using filter-while-typing, just search for `/` - unless you have a very weird database structure, that should match every entry you have.

maximbaz:
@erayd found a tiny nitpick, not worth wasting time if the fix is not obvious, but maybe it is obvious to you?

Open browserpass, it is in filter mode, start typing something until no results appear, extension switches to the search mode, but the bottom bar with "No matching passwords for github.com" appears. Ideally it should not appear, just like if you press Backspace immediately after opening browserpass.

erayd:
@maximbaz That was deliberate - i.e. "you have filtered everything, and there are no longer any results that match."

I can change it though, if you think that some other behaviour would be better?

maximbaz:
Aaah, no, it makes total sense, it's just not instantly obvious that when you are in this state, you are at the same time in the search mode (you see, I've actually asked about this already...).

Let's keep it as it is now.

Issue #164, opened 2017-09-29
Report meaningful errors, not "Error talking to browserpass"


maximbaz:
We are sick of seeing "Error talking to browserpass". Time to report meaningful errors.

--------------

UPDATE: since 2.0.18 we will have errors from the host app printed in stdout and sent to the extension. 

What needs to be done next:

- In addition to stdout, log the errors to a file, e.g. `/var/log/browserpass.log`
- Test the errors, make sure what is shown in the extension is not very cryptic.
- Create a standardized message structure for communication between extension and the host app, have separate fields for the error details and successful results.

----------

As a very first step, we should introduce logging for the host app. This way we can at least see if it was triggered or not and which actions it took.

Next step is passing error messages back to the browser. I was playing with this a little bit, it would require some refactoring. Today browserpass host app crashes on any error, and prints details on stderr. However Chrome doesn't read stderr, it only reads stdout. Also, it doesn't like crashes, on unexpected crash it prints this famous "Error talking to browserpass".

If we are going to change this, the approach would be to prevent the host app from crashing, like ever, and instead send the error message back to browser. We could come up with some standardized JSON structure, e.g. `{"status": "OK/FAIL", "messsage": "<message from host app" }` to distinguish errors from correct responses.

It is also worth keeping in mind https://github.com/dannyvankooten/browserpass/issues/20#issuecomment-297159280, as it suggests to bring this idea one step further and not only never crash the host app, but actually never even stop the connection between browser and the app. Today the connection is established and closed as needed, instead we could establish one connection and keep it open forever. Such always open connection could maybe even help with https://github.com/dannyvankooten/browserpass/issues/77#issuecomment-332673549.

One more interesting idea is to check the versions - in case of an error, among error details the host app should send its version back to the extension, so that we can additionally do a version check and suggest users to update. It is a bit tricky, because sometimes we have no changes in host app and browser extension is actually expected to work with previous version of the host app.

It doesn't have to be implemented all at once, I'm just throwing ideas in and connecting the dots 🙂

chrboe:
`chrome.runtime.connectNative` would be a good idea in general. I know it's really not relevant for our use case, but the systems programmer in me cringes at the thought of spawning a new process for every request 🙂

jwiegley:
[...]

PR #224, opened 2018-03-20
Add feature to launch URLs & support basic auth

### LINKED ISSUE #103: Support HTTP basic authentication (browser pop-up window)
I'm very happy with browserpass+GnuPG and removed all my insecure FF credential storage.

There are two issue (I will file the 2nd one in an extra issue): Some sites use for the credential an extra pop-up window and in this case one can not hit the browserpass icon anymore. It helps only to cut&paste from another (shell) window. Perhaps there is no solution for this.

### LINKED ISSUE #214: Add a button to open entry in a new tab (and use `url` field if it exists)
### General information
Would it be possible to add a feature to read `url` field?
I liked this in LastPass where you just search through your items, click it and the browser opens your stored URL.

* Operating system + version: Manjaro KDE
* Browser + version: Waterfox 56.0.4.1
* Browserpass extension version as reported by your browser: 2.0.11

erayd:
- Host app will now return a "url" field
 - Browser plugin has a "launch URL" button for each entry
 - Browser plugin has a new hotkey (g) to launch the URL
 - Supply basic auth credentials if necessary when launching via the extension

Note that basic auth will only work in Firefox >= 54. Older versions don't have the required API to do this.

Closes #214.
Closes #103.


maximbaz:
Nice! Let me throw some ideas first.

1. Could you please [check this out](https://github.com/dannyvankooten/browserpass/issues/103#issuecomment-313845253)? When we are opening a new tab, can we try to additionally override `onAuthRequired` and prefill the user credentials, so that user is already logged in? As a bonus, this will also solve #103.

1. When host app does not return a `url`, instead of just saying "sorry we can't do it", can we open the url of the entry name itself, if it looks like a domain? In other words, if my file name is `google.com.gpg`, I'd like browserpass to open google without having to add `url: https://google.com` to the file. It's also not nice to render the button to open URL, when it cannot actually open the URL (unless I add this `url: ` metadata to each and every password entry).

erayd:
@maximbaz 
1. Will take a look and see what I can do.

2. I will add this. Note that this will require bundling a TLD list with the extension - simply guessing which path component is the URL doesn't feel like a good idea (because it could result in things like `http://work` or `http://myUsername`).

maximbaz:
Thanks! Hmm, you are right, it is not very clear how to select the domain part from the entry name... 

Although I wouldn't suggest building a super comprehensive domain matcher, because the workaround is very simple (adding `url:` to the password entry) - it should just work in most common cases.

erayd:
@maximbaz I've implemented extracting a URL from the entry path. What do you think?

erayd:
@maximbaz HTTP basic auth now works too :-).

chrome/script.browserify.js
```diff
@@ -232,6 +236,47 @@ function getFaviconUrl(domain) {
...
+function launchURL() {
+  var what = this.what;
+  var entry = this.entry;
+  chrome.runtime.sendNativeMessage(
+    app,
+    { action: "get", entry: this.entry },
+    function(response) {
+      if (chrome.runtime.lastError) {
+        error = chrome.runtime.lastError.message;
+        m.redraw();
```
maximbaz: nitpick: could you please reduce the nesting? 
It's a little difficult to see which "else" belongs to which "if" 🙂

```
if (error) {
  ...
  return; 
}
```

chrome/script.browserify.js
```diff
@@ -232,6 +236,47 @@ function getFaviconUrl(domain) {
...
+      } else {
+        // get url from login path if not available in the host app response
+        if (!response.hasOwnProperty("url") || response.url.length == 0) {
+          var parts = entry.split(/\//).reverse();
```
maximbaz: I see pros and cons in reversing the parts, and so far I found only one case among my passwords where this is not ideal. We will probably keep this as it is, but I want to ask anyway, what is your reasoning for and against reversing?

chrome/script.browserify.js
```diff
@@ -232,6 +236,47 @@ function getFaviconUrl(domain) {
...
+              response.url = part;
+              break;
+            }
+          }
+        }
+        // if a url is present, then launch a new tab via the background script
+        if (response.hasOwnProperty("url") && response.url.length > 0) {
+          var url = response.url.match(/^([a-z]+:)?\/\//i) ? response.url : "http://" + response.url;
+          chrome.runtime.sendMessage({action: "launch", url: url, username: response.u, password: response.p});
+          window.close();
```
maximbaz: here's another potentially good place to do `return` and reduce nesting.

chrome/script.browserify.js
```diff
@@ -232,6 +236,47 @@ function getFaviconUrl(domain) {
...
+        } else {
+          // no url available
+          if (!response.hasOwnProperty("url")) {
+            resetWithError("Your host application is too old - must be at least 2.0.14 for URL launch.");
```
maximbaz: First brick in building #164, I like this 👍 

However, we need to adjust the message, because with added parsing of domain it's not consistent anymore. With the old host application I can open url for password entries like "personal/google.com" just fine, until one day I try to open url for the entry "personal/notes" and suddenly get the error "Your host application is too old". 

I think we should say something like this: 

Unable to detect the URL to open for this entry. If you defined it in the password file, update your host application to the latest version.

chrome/script.browserify.js
```diff
@@ -232,6 +236,47 @@ function getFaviconUrl(domain) {
...
+          } else {
+            resetWithError("No URL is available for this login.");
```
maximbaz: In my mind something like this will sound better, maybe there is a URL but it's just us who cannot detect it:

Unable to detect the URL to open for this entry.

maximbaz:
Attaching the compiled extension for the current code for others to test:

[chrome-browserpass.crx.zip](https://github.com/dannyvankooten/browserpass/files/1833193/chrome-browserpass.crx.zip)

chrome/background.js
```diff
@@ -95,6 +95,42 @@ function onMessage(request, sender, sendResponse) {
...
+        function authListener(requestDetails) {
+          // remove event listeners once tab loading is complete
+          chrome.tabs.onUpdated.addListener(function statusListener(tabId, info) {
+            if (info.status === "complete") {
+              chrome.tabs.onUpdated.removeListener(statusListener);
+              chrome.webRequest.onAuthRequired.removeListener(authListener);
+            }
+          });
+          // only supply credentials if this is the first time for this tab
+          if (!authAttempted) {
```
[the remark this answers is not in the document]
erayd: 1. Create a new tab, and get the tab ID in the callback
2. Add onAuthRequired listener
3. When the auth callback fires, add a status listener to the tab
4. Supply login credentials if we haven't already supplied them
5. When the status listener fires and the status is complete (i.e. the page has loaded), remove both listeners.

The reason both the `authCompleted` and the listeners are required:
 * The `onAuthRequired` listener cannot remove itself, as doing so crashes the extension.
 * Because it cannot remove itself, it may fire multiple times (e.g. if the password is incorrect)
 * So the boolean restricts it to one time only
 * And removing it in the status listener once the page is complete ensures that the event is no longer live, and avoids a crash.

> is it guaranteed that `info.status` will become complete only after basic auth is filled in?

I could not find any specific documentation guaranteeing this, and I looked pretty hard. However, the documentation does say that this occurs when the page load is complete, and blocking authentication requirements prevent the page from loading until they are satisfied. I've tested it extensively, and I'm comfortable that this does seem to be the case, but I cannot absolutely guarantee that it is so.

chrome/background.js
```diff
@@ -95,6 +95,42 @@ function onMessage(request, sender, sendResponse) {
...
+                ;
+              if (!confirm(message)) {
+                return {};
+              }
+            }
+            return {authCredentials: {username: request.username, password: request.password}};
+          }
+        },
+        {urls: ["*://*/*"], tabId: tab.id},
+        ["blocking"]
```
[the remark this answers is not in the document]
erayd: It's not possible to mess with the request asynchronously (including providing authentication details). The `"blocking"` option ensures that the request waits for our callback to complete before proceeding.

chrome/manifest.json
```diff
@@ -7,7 +7,7 @@
   "author": "Danny van Kooten",
   "homepage_url": "https://github.com/dannyvankooten/browserpass",
   "background": {
-    "persistent": false,
+    "persistent": true,
```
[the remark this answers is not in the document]
erayd: Chrome doesn't allow non-persistent event pages to use the `webRequest` API. If you have this set to `false`, you'll get a massive warning when you try to load the extension telling you that this isn't possible, and the Chrome will refuse to load the extension.

chrome/script.browserify.js
erayd: Sure.

chrome/script.browserify.js
erayd: It seems more likely that someone will have a workplace that is a valid domain (e.g. "I work for example.com"), than a username that is a valid domain. It also seems more likely that if domains are nested, the deepest one is the most likely to be what the user wants. So reversing it prioritises domains that are found closer to the end of the entry.

chrome/script.browserify.js
erayd: Done.

chrome/script.browserify.js
erayd: Good point - done.

chrome/background.js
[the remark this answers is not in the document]
maximbaz: thank you, makes sense!

chrome/background.js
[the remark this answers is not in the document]
maximbaz: I see, thanks!

chrome/script.browserify.js
maximbaz: Agreed

chrome/background.js
```diff
@@ -95,6 +95,42 @@ function onMessage(request, sender, sendResponse) {
...
   }
+
+  // spawn a new tab with pre-provided credentials
+  if (request.action == "launch") {
+    chrome.tabs.create({url: request.url}, function (tab) {
+      var authAttempted = false;
+      chrome.webRequest.onAuthRequired.addListener(
+        function authListener(requestDetails) {
+          // only supply credentials if this is the first time for this tab
+          if (!authAttempted) {
```
maximbaz: could you invert this too, i.e. `if (authAttempted) { return {}; }`? Also judging by line 123 returning `{}` is important in this case too.

erayd: No problem - done :-).

maximbaz:
Just as a final confirmation, do you want to do anything else here or can I merge?

erayd:
> Thanks once again, this is awesome!

Most welcome :-).

> Just as a final confirmation, do you want to do anything else here or can I merge?

If you could give me two minutes, I want to tweak the 'insecure connection' error message slightly. Stand by.

erayd:
@maximbaz Thanks - I just wanted to clarify that message a bit; I figured the original wording wasn't the clearest.

I'm happy with it now - all good to merge :-). Thanks very much for your help with this, and your extremely speedy code reviews!

chrome/background.js
```diff
@@ -95,6 +95,44 @@ function onMessage(request, sender, sendResponse) {
...
+      chrome.webRequest.onAuthRequired.addListener(
+        function authListener(requestDetails) {
+          // only supply credentials if this is the first time for this tab
+          if (authAttempted) {
+            return {};
+          }
+          authAttempted = true;
+          // remove event listeners once tab loading is complete
+          chrome.tabs.onUpdated.addListener(function statusListener(tabId, info) {
+              if (info.status === "complete") {
```
maximbaz: @erayd do you think this code should also compare `tabId` with the `tab.id` before removing the listeners?

erayd: It doesn't need to, because the listeners are specific to the tab anyway.

There is a security problem, but that's not it. I'm filing an issue now.

maximbaz: Are you sure that this doesn't register a listener that fires for _all tabs_? Maybe it's just a weird API, but the tab object is not used when registering the listener, and the `tabId` is provided as a first argument for a reason, both of these lead me to think that this will fire for all tabs...

I know that this is not about the security problem, it's just what I accidentally stumbled upon 🙂

erayd: Ooh, good point. The `onAuthRequested` handler is specific to the tab, but the status listener is not. My bad.

I'm going to roll the security fix PR right now - I feel quite bad I didn't notice that in the first place; I'm normally more call than that. I will fix this at the same time.

maximbaz: Thank you!

uri:
Aside from this feature addressing the basic auth issue #103, what is the purpose of this? It seems strange to have this on all pass entries. Personally I don't have a single basic auth entry, so for me the button is a worse way to navigate to url? Is there another common use case for this that I'm missing?

maximbaz:
See #214 for more details, people seem to like this way of opening websites in Lastpass and other password managers, and they carry their habits to browserpass.

In addition it now allows you to use `url:` metadata in your pass files. For example, you might have a file called `github.com.gpg`, but inside of it you add `url: https://github.com/login`, and thus when you open github via browserpass, it opens the login form if you are unauthorized, or the home page if you are already authorized.

uri:
>In addition it now allows you to use url:

Oh now that is useful.

#235

PR #226, opened 2018-03-20
Allow using a bare TOTP token instead of an otpauth:// URL

erayd:
If an `otpauth://` URL is not present, but `totp:` or `totp-secret:` is available, then use that with the default settings for generating an OTP code.

### DIFF
--- a/browserpass.go
+++ b/browserpass.go
@@ -179,9 +179,16 @@ func readLoginGPG(r io.Reader) (*Login, error) {
 }
 
 func parseTotp(str string, l *Login) error {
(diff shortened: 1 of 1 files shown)


maximbaz:
I don't use OTP via pass so I can't really test this, but the change looks all right to me.

Let me ping @qbit who implemented the OTP support, if he doesn't have objections until the next release, I'll merge.

qbit:
This is more in line with how it's done in [gopass](https://github.com/justwatchcom/gopass/blob/master/pkg/otp/otp.go#L29). Tests fine for me.

PR #231, opened 2018-03-22
[SECURITY] Fix for credential leak vector #230

### LINKED ISSUE #230: [SECURITY] Credential leak vector
#224 introduces a credential leak via HTTP basic authentication.

In order for credentials to leak, **all** of the following conditions must be true:
 * The tab must be launched via the browserpass extension;
 * The tab must have never requested any kind of modal (e.g. basic) authentication;
 * The first request for modal authentication in the tab must be malicious;
 * The modal authentication request must occur via HTTPS.

If all those are true, then the credentials of the launched site will be invisibly provided to the modal authentication request.

This scenario can occur in two ways:
 1. A site is launched which does not require modal authentication; or
 2. A site is launched which *does* require modal authentication, but has already been authenticated, so the auth request never occurs.

I apologise for this oversight - this is my fault. I will submit a PR that fully closes this vector ASAP, and within the next 24 hours.

In the meantime, in order to avoid the vulnerability, users should not launch sites via the browserpass extension unless they know that a basic auth request will occur before they navigate away from the site, or they should ensure that any navigation away from a browserpass-launched site occurs in a different tab.

erayd:
This PR fixes #230:
 * Credentials are now discarded immediately as soon as the tab has loaded
 * Credentials are not supplied to any model login that occurs after the tab has loaded
 * If the domain requesting credentials is not the same as the domain that was launched, the user will be asked if they really want this.

Once again, I apologise for #230 - that is my fault, and my oversight. I'm usually more careful than that, and I sincerely regret that I allowed this one to slip through.

### DIFF
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -107,9 +107,11 @@ function onMessage(request, sender, sendResponse) {
         tabId,
         info
       ) {
-        chrome.tabs.onUpdated.removeListener(tabCompleteListener);
-        tabLoaded = true;
-        request.username = request.password = null;
+        if (tabId == tab.id && info.status == "complete") {
+          chrome.tabs.onUpdated.removeListener(tabCompleteListener);
+          tabLoaded = true;
+          request.username = request.password = null;
+        }
       });
 
       // intercept requests for authentication
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -102,83 +102,79 @@ function onMessage(request, sender, sendResponse) {
       var tabLoaded = false;
       var authAttempted = false;
 
+      // listener function for authentication interception
+      function authListener(requestDetails) {
+        // only supply credentials if this is the first time for this tab, and the tab is not loaded
+        if (authAttempted) {
+          return {};
+        }
+        authAttempted = true;
+
+        // don't supply credentials for loaded tabs
+        if (tabLoaded) {
+          return {};
+        }
+
+        // ask the user before sending credentials to a different domain
+        var launchHost = request.url.match(/:\/\/([^\/]+)/)[1];
+        if (launchHost !== requestDetails.challenger.host) {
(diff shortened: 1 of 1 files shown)


erayd:
@maximbaz Any chance of a code review? I'd rather this was merged sooner than later to get that hole closed off ASAP.

maximbaz:
> Credentials are now discarded immediately as soon as the tab has loaded

So now we intentionally make it so that browserpass will never fill credentials on https://www.httpwatch.com/httpgallery/authentication/ after clicking on "Display image", even if this page was opened via browserpass. Is that right?

chrome/background.js
```diff
@@ -99,46 +99,82 @@ function onMessage(request, sender, sendResponse) {
...
+
+      // intercept requests for authentication
+      chrome.webRequest.onAuthRequired.addListener(
+        authListener,
         { urls: ["*://*/*"], tabId: tab.id },
         ["blocking"]
       );
+
+      // notice when the tab has been loaded
+      chrome.tabs.onUpdated.addListener(function tabCompleteListener(
```
maximbaz: I have a different proposal for this, but I cannot do it now. Let me merge this as it is and publish a new version, and I'll make a separate PR and ping you for feedback.

erayd:
> So now we intentionally make it so that browserpass will never fill credentials on https://www.httpwatch.com/httpgallery/authentication/ after clicking on "Display image", even if this page was opened via browserpass. Is that right?

Yes.

Going forward, I'd like to address this use-case by intercepting the auth request and allowing the user to choose a credential, but I need to think about how to do this in a sane manner. In the meantime, simply disallowing the scenario feels much safer.

chrome/background.js
erayd: OK.

PR #232, opened 2018-03-22
Added titles to buttons for a better user experience

AuspeXeu:
The title says is already :)

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -34,9 +34,10 @@ function view() {
         m.trust(`No matching passwords found for <strong>${domain}</strong>.`)
       );
     } else if (logins.length > 0) {
+      const autoSubmit = localStorage.getItem("autoSubmit") == "true";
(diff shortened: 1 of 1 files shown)


chrome/script.browserify.js
```diff
@@ -36,7 +36,7 @@ function view() {
     } else if (logins.length > 0) {
       results = logins.map(function(login) {
         let selector = "button.login";
-        let options = { onclick: getLoginData.bind(login) };
+        let options = { onclick: getLoginData.bind(login), title: "Fill form" + (localStorage.getItem("autoSubmit") ? " and submit" : "") };
```
maximbaz: could you change the code to access localStorage only once and store result in the variable, before the `logins.map()`?

Also, this doesn't really work if you change autosubmit to "yes", save, and then change back to "no".
It should be instead `localStorage.getItem("autoSubmit") == "true"`

PR #233, opened 2018-03-22
Refactor basic auth

maximbaz:
- Use a shared tabs.onUpdated instead of subscribing multiple times
- Extract onAuthRequired into a separate global function

@erayd I'll comment inline on a couple of particularly note-worthy points.

### DIFF
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -3,6 +3,7 @@
 var app = "com.dannyvankooten.browserpass";
 
 var tabInfos = {};
+var authListeners = {};
 
 chrome.runtime.onMessage.addListener(onMessage);
 chrome.tabs.onUpdated.addListener(onTabUpdated);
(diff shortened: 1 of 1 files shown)


chrome/background.js
```diff
@@ -99,96 +100,86 @@ function onMessage(request, sender, sendResponse) {
   // spawn a new tab with pre-provided credentials
   if (request.action == "launch") {
     chrome.tabs.create({ url: request.url }, function(tab) {
-      var tabLoaded = false;
```
maximbaz: I didn't see a particular use for `tabLoaded`, given that we discard onAuthListener immediately after tab is loaded. Did you have a point for this that I'm missing?

chrome/background.js
```diff
@@ -99,96 +100,86 @@ function onMessage(request, sender, sendResponse) {
...
-        info
-      ) {
-        if (tabId == tab.id && info.status == "complete") {
-          // remove listeners
-          chrome.tabs.onUpdated.removeListener(tabCompleteListener);
-          chrome.webRequest.onAuthRequired.removeListener(authListener);
-
-          // mark tab as loaded & wipe credentials
-          tabLoaded = true;
-          request.username = request.password = null;
```
maximbaz: The reference to this object is in this function and now in `onAuthRequired`, as soon as both functions are gone the credentials are not accessible anyway - is there a point in additionally wiping them here?

chrome/background.js
erayd: That can go. I went through a couple of different versions of the bugfix; that was left over from one of them but is no longer applicable.

chrome/background.js
erayd: As above - this was leftover from an earlier iteration of the bugfix; it can go.

maximbaz:
Thanks @erayd for the confirmations, any other feedback before this is merged?

erayd:
@maximbaz No other feedback at this point - I like what you've done, go ahead and merge it :-).

PR #235, opened 2018-03-23
Update README.md to include launch URL info

uri:


maximbaz:
Thanks! Regarding the metadata, there are actually more fields supported, do you think it is worth mentioning them all?

https://github.com/dannyvankooten/browserpass/blob/604f560e390e56add8ea751c5928ca08fe32ce5a/browserpass.go#L217

maximbaz:
I decided to mention them all, thanks once again 😉

uri:
Good to know, thanks!

PR #237, opened 2018-03-24
Allow configuring multiple custom password stores

### LINKED ISSUE #34: Search passwords only in a specified subfolder of my password-store

### LINKED ISSUE #77: Support gopass mounts

### LINKED ISSUE #98: Option to specify password store directory

maximbaz:
I had a few hours free, so I decided to play with implementing this feature.

I suggest reviewing the diff [without whitespace changes](https://github.com/dannyvankooten/browserpass/pull/237/files?w=1).

The idea is to allow configuring multiple password stores in options, and even toggle them when needed:

--------

image

image

--------

image

image

---------

A folder can be a path to a custom password store (#98), a path to a subdirectory of a password store (#34) or a path to gopass mount (#77).

## Known limitations

Nothing left here 🙂

<details>
<summary>Fixed</summary>
1. It has to be an absolute path, no expansion of `~` or environment variables.
</details>

<details>
<summary>Fixed by @erayd in #239</summary>

1. I didn't have time (and desire) to dynamically add/remove more lines for directories, so I pre-created 5 empty input lines. Because 5 should be enough for everyone 😄 I'm joking, but if you want to implement something fancier, please submit a new PR.

1. Current solution doesn't support having files with identical names. In other words, if you enable two password stores in the options:

```
/home/user/.password-store-1/
/home/user/.password-store-2/
```

And you have identically named files in both password stores, like this:

```
/home/user/.password-store-1/github.com.gpg
/home/user/.password-store-2/github.com.gpg
```

Only the first `github.com.gpg` will be seen by browserpass. Use checkbox in the extension options to temporarily disable password store that you don't need. 

However if the files are in different subfolders (like below), then it's fine:

```
/home/user/.password-store-1/personal/github.com.gpg
/home/user/.password-store-2/work/github.com.gpg
```
</details>

----

Fixes #98 
Fixes #34 
Fixes #77


erayd:
> Current solution doesn't support having files with identical names.

What's the limitation causing this? Would adding a temporary prefix be a suitable workaround?

pass/disk.go
```diff
@@ -90,24 +92,33 @@ func (s *diskStore) GlobSearch(query string) ([]string, error) {
...
 		if err != nil {
 			return nil, err
 		}
-		items[i] = strings.TrimSuffix(item, ".gpg")
+
+		allMatches := append(matches, matches2...)
+
+		for i, match := range allMatches {
+			// TODO this does not handle identical file names in multiple s.paths
+			item, err := filepath.Rel(path, match)
```
maximbaz: @erayd this is the cause of the limitation, it generates the entry name by cutting off the configured prefix. A better solution would need to account for multiple configured paths to make sure it cuts off as much as possible, but still keeping the resulting string unique among all paths.

pass/disk.go
```diff
@@ -125,17 +136,16 @@ func (s *diskStore) GlobSearch(query string) ([]string, error) {
...
-	if os.IsNotExist(err) {
-		return nil, ErrNotFound
+	for _, path := range s.paths {
+		path := filepath.Join(path, item+".gpg")
+		f, err := os.Open(path)
+		if os.IsNotExist(err) {
+			continue
+		}
+		// TODO this does not handle identical file names in multiple s.paths
+		return f, err
```
maximbaz: And then here, when we receive an entry "github.com", the code is trying to guess, which password store it belongs to. Right now the code is simple, the first password store that contains "github.com.gpg" is the one 🙂

pass/disk.go
erayd: Mind if I open a PR against this branch with a proposed workaround?

maximbaz: Please go ahead 👍

erayd: Done - see #239.

maximbaz:
Description updated, all limitations fixed.

maximbaz:
> Looks like this picked up a lot that prettier did not notice when I ran it via docker. Is the makefile config for running prettier via docker correct?

Mmm, the makefile is the same and I just ran `make` that should be run by docker. I haven't been using this docker image for a while though... When you use the docker image, does it fix any kind of style issues at all? No errors printed? Maybe try to rebuild the docker container?

This is the output when I run `make` by the way, I think when you use docker image it should print exactly the same output.

image

erayd:
That looks quite different - I just get this:
scr-20180327-223340

pass/disk.go
```diff
@@ -14,100 +14,126 @@ import (
...
+	entries, err := s.GlobSearch("")
 	if err != nil {
 		return nil, err
 	}
 
+	// GlobSearch now results `storename:filename`, for fuzzy search we need to provide only file names
+	var fileNames []string
+	for _, entry := range entries {
+		fileNames = append(fileNames, strings.SplitN(entry, ":", 2)[1])
+	}
```
maximbaz: @erayd you'll love this 🙂 Caught by a unit test, that suddenly started finding 10 entries instead of 1 for a search term `"def"` (which started to match store name "default")

erayd: Nice catch :-).

I actually use the store name as part of my search fairly commonly, to distinguish between files in different stores that have otherwise identical names. However, I don't actually need that capability in the host app - having it in the "refine search" filter (which also includes the store name in searches) is enough for my use-case :-).

maximbaz: Good point, until now I didn't even think to test how "refine search" behaves with store names, but I agree this is a very common use-case to refine by the store name. In fact, I'm used to that too, because I have both "work" and "personal" in the same password store, but now I can configure different browser profiles to use different subfolders and this reduces 99% of cases where I needed to refine - now I just press Enter 🙂

PR #239, opened 2018-03-27
Rewrite to allow matches from multiple stores

erayd:
* Rewrite the multiple-store feature to allow files with the same name in different stores
 * Add dynamic number of stores
 * Rewrite options screen
 * Add styling to login entries when multiple stores are enabled
 * Various other minor improvements.


erayd:
@maximbaz Worth clarifying, because this may not be obvious - the browser extension in this branch is backwards-compatible with the current release version of the host app. There are no breaking changes.

erayd:
There's also something in this PR which is causing Firefox to occasionally display a horizontal scrollbar. It's not the `min-width` of the popup. Will update when I've found the cause.

erayd:
Firefox scroll issues sorted.

maximbaz:
This looks really nice! I particularly like that the label is not displayed when only one password store is configured.

chrome/options.browserify.js
```diff
@@ -0,0 +1,120 @@
...
+    value: false
+  },
+  use_fuzzy_search: {
+    type: "checkbox",
+    title: "Use fuzzy search",
+    value: true
+  },
+  customStore: {
+    title: "Custom password store locations",
+    value: [{name: "one", enabled: true, name: "fish", path: "/home/fish"}],
```
maximbaz: This default entry "fish" is confusing, the first time I saw it I toggled the checkbox twice, closed, and browserpass started showing me error `Error: lstat /home/fish: no such file or directory`. I will change this line to `value: [{}]` if you don't have objections - the placeholders explain well what needs to be entered where.

pass/disk.go
```diff
@@ -109,11 +114,11 @@ func (s *diskStore) GlobSearch(query string) ([]string, error) {
 
 		for i, match := range allMatches {
 			// TODO this does not handle identical file names in multiple s.paths
```
maximbaz: Now you can remove the TODO, here and below in `Open` 🙂

pass/disk.go
```diff
@@ -14,46 +14,51 @@ import (
 	sfuzzy "github.com/sahilm/fuzzy"
 )
 
+// StoreDefinition defines a password store object
+type StoreDefinition struct {
```
maximbaz: Could you please also adjust `pass/disk_test.go` for these changes?

browserpass.go
```diff
@@ -33,10 +33,12 @@ var endianness = binary.LittleEndian
...
 // which options have been selected by the user, and put them in a JSON object
 // which is then passed along with the command over the native messaging api.
+
+// Config defines the root config structure sent from the browser extension
 type Config struct {
 	// Manual searches use FuzzySearch if true, GlobSearch otherwise
-	UseFuzzy bool     `json:"use_fuzzy_search"`
-	Paths    []string `json:"paths"`
+	UseFuzzy    bool                   `json:"use_fuzzy_search"`
+	CustomStore []pass.StoreDefinition `json:"customStore"`
```
maximbaz: should we maybe call it in plural?

chrome/options.browserify.js
erayd: Oops - that was a leftover test value. I've just nuked it. Thanks for picking it up :-).

pass/disk.go
erayd: Done.

browserpass.go
erayd: Could you clarify what you mean by _"call it in plural"_ please? I'm not sure what you are asking me to do.

maximbaz: sorry, I meant name it "CustomStores" 🙂

pass/disk.go
erayd: Yes. I just need to figure out what needs doing - I don't work much with Go, so I'm a bit slow with it.

chrome/options.browserify.js
erayd: (To clarify - the default is supposed to be an empty array; it doesn't need to contain a value).

browserpass.go
```diff
@@ -47,6 +49,18 @@ type msg struct {
 	Entry    string `json:"entry"`
 }
 
+func SendError(err error, stdout io.Writer) error {
```
maximbaz: My special THANK YOU for this!

I'm going to separately test this more, but what is your opinion on this as a fix for #164? I personally want to see errors logged and reported back to the extension, this code already sends the message back and logs it to stdout, I would only add logging to a file too (not necessarily in this PR). Anything you would additionally want for better error reporting?

chrome/options.browserify.js
maximbaz: I tried a few options before writing that comment, and actually I found it quite pleasant for the eye to have one empty line pre-created

browserpass.go
erayd: Thanks for clarifying - done :-).

pass/disk.go
maximbaz: I can fix the tests myself after merging too, that's not a big deal 🙂

browserpass.go
erayd: > My special THANK YOU for this!

You're most welcome :-).

I think this goes partway to addressing #164, but I don't consider it a full fix.

I added it because I needed a way to get error info out of the host app while I was debugging it, and figured I'd do so in a way that was more generally useful. So it helps - but I don't think that #164 should be closed because of it. I suspect there's still a fair bit of error information that may not be caught.

erayd: Besides, my way of doing it here is a little messy ;-). Anything that's returned as a raw string is considered an error. In the longer term, I'd like to see the return message standardised properly, and include a status code. But I couldn't do that without breaking backwards-compatibility, hence the string thing.

maximbaz: I had that idea too, but then I saw your approach with strings and I thought that perhaps this is actually fine, we can log more details in the log file (detailed messages, error code, etc), and have the extension receive just a string error. That way we don't break compatibility, and it's easy to add to ISSUE_TEMPLATE an ask to print the last 100 lines from `/var/log/browserpass.log` 🙂

chrome/options.browserify.js
erayd: Makes sense - I've added your suggested default in f686716.

pass/disk.go
erayd: That works too :-). If you're happy to do that, I would definitely prefer it - you are likely to pick up things in testing that I may miss due to my lack of familarity.

browserpass.go
erayd: It's not really fine in the longer term though, because at the moment we can't change much functionality-wise without breaking stuff (and besides, co-opting the entire `string` datatype for errors isn't very nice, because nobody can use it for anything else). Switching to a consistent and extensible message format will allow much more scope to add functionality without requiring a host app update every time we do something significant.

I don't think this PR is the right place for such a change, but IMO it definitely needs to happen at some point.

maximbaz: Agreed, and I will update the ticket with your comment :)

PR #240, opened 2018-03-27
Fix prettier when running from docker

maximbaz:
`yarn run` for some reason does not pass `--write` to the prettier when executed from the docker container.

So instead of #236 I will do this.

/cc @belak, @erayd


belak:
The other option would be to include the whole command in the scripts. `yarn run prettier-css` as an example.

erayd:
@maximbaz Thanks!

PR #247, opened 2018-04-03
Use store path as hover title for store badge

erayd:
When hovering over the store name badge in multi-store mode, display the path to that store.

scr-20180403-150730

### DIFF
--- a/chrome/script.browserify.js
+++ b/chrome/script.browserify.js
@@ -49,11 +49,18 @@ function view() {
         };
 
         var store = "default";
+        var storeTitle = "Default password store";
         var name = login;
         var i;
         if ((i = login.indexOf(":"))) {
           if (searchSettings && searchSettings.customStores.length > 1) {
             store = login.substr(0, i);
+            searchSettings.customStores.map(function(customStore) {
+              if (customStore.name != store) {
+                return;
+              }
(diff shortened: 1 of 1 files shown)


chrome/script.browserify.js
```diff
@@ -49,11 +49,18 @@ function view() {
         };
 
         var store = "default";
+        var storeTitle = "Default password store";
         var name = login;
         var i;
         if ((i = login.indexOf(":"))) {
           if (searchSettings && searchSettings.customStores.length > 1) {
             store = login.substr(0, i);
+            searchSettings.customStores.map(function(customStore) {
```
maximbaz: unusual use of `map` 🙂

erayd: Oops - that was a mistake. Intention was `forEach`. I think I must have been looking at `map` elsewhere, and typed it accidentally.

No real functional impact using it here though - so we can leave it as-is - but would you like me to change it for clarity?

erayd: Or I guess I could have used a `(for i in searchSettings.customStores) {` loop too.

maximbaz: I'll maybe change to `forEach` next time I need to do some changes in that file, as you said there is no functional impact anyway 😉

PR #252, opened 2018-04-05
Add per-site autosubmit

### LINKED ISSUE #251: Support per-site configuration for autoSubmit in pass entries
Hello, thank you for this nice peace of SW :) 

I have one suggestion about auto-submit. Currently is possible to have it enabled for all websites or disabled for all websites, unfortunately I found one login page with captcha and this one I could not submit before is filled in. For this reason I need to disallow autosubmit for all websites.  

my expectation is: in pass multiline option to force global extension setting I fill up line like  
```
autoSubmit: false
```

erayd:
Adds per-site autosubmit by allowing `autosubmit: <true|false>` in password files to override the global config option.

Requires a host app update to work, but using this version of the extension with the old host app will degrade gracefully.

Closes #251.

### DIFF
--- a/chrome/background.browserify.js
+++ b/chrome/background.browserify.js
@@ -76,7 +76,7 @@ function onMessage(request, sender, sendResponse) {
           ) {
             // do not send login data to page if URL changed during search.
             if (tabs[0].url == request.urlDuringSearch) {
-              var autoSubmit = response.asOverride ? response.autoSubmit : getSettings().autoSubmit;
+              var autoSubmit = response.hasOwnProperty('autoSubmit') ? response.autoSubmit : getSettings().autoSubmit;
               fillLoginForm(response, tabs[0], autoSubmit);
             }
           });
--- a/chrome/background.browserify.js
+++ b/chrome/background.browserify.js
@@ -12,7 +12,7 @@ chrome.tabs.onUpdated.addListener(onTabUpdated);
 chrome.runtime.onInstalled.addListener(onExtensionInstalled);
 
 // fill login form & submit
-function fillLoginForm(login, tab) {
+function fillLoginForm(login, tab, autoSubmit) {
   const loginParam = JSON.stringify(login);
   chrome.tabs.executeScript(
     tab.id,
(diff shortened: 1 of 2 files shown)


chrome/background.browserify.js
```diff
@@ -76,7 +76,8 @@ function onMessage(request, sender, sendResponse) {
           ) {
             // do not send login data to page if URL changed during search.
             if (tabs[0].url == request.urlDuringSearch) {
-              fillLoginForm(response, tabs[0]);
+              var autoSubmit = response.asOverride ? response.autoSubmit : getSettings().autoSubmit;
```
maximbaz: why not simply check if `response.autoSubmit` is not undefined instead of introducing `asOverride`?

erayd: Because go doesn't leave it undefined; it defaults to `false`.

Without `asOverride` there is no way to distinguish between a default `false`, and an explicitly specified `false` that should override the global setting.

maximbaz: Let's try other approaches first before using `asOverride`, because it feels like a terrible hack...

First, you could try to make `autoSubmit` of a type `*bool`, a pointer to bool. That way the pointer can be `nil` (which is undefined) or `true/false`. I'm not sure how that would work with serialization, would it be transformed to `undefined` or not.

Another idea is to make `autoSubmit` a string, where you can use empty string as an indication of "value is not set".

Also, you can omit empty values completely during serialization, mark them as:

```
AutoSubmit string   `json:"autoSubmit,omitempty"`
```

erayd: I had no idea `omitEMpty` was a thing - that's **much** nicer! I'll do that now, stand by...

maximbaz: I'm not sure it will work for `bool` type though, but should work for `*bool`...

erayd: How's that?

maximbaz: With this approach, can you enable the auto-submit globally and override it with `autoSubmit: false` for one website?

I'm just worried that with `bool` type the `omitempty` considers `false` as empty, regardless of whether `false` is the default value or was set by you.

With `*bool` type, I would expect `omitempty` to understand that `nil` value is the empty value, while `true` and `false` are values that should not be omitted.

erayd: `*bool` fixed it :-).

erayd: That was a great idea of yours by the way - thank you :+1:.

maximbaz: Thanks to you for implementing 😉  I'll test both PRs in the evening

Issue #267, opened 2018-06-24
Navigate between password/login/otp using arrow keys


rbuzatu90:
I think using arrow keys to select the desired field is way more intuitive then using ctrl/shift+c.
LastPass has this option and is very convenient.

rbuzatu90:
If there is someone who's willing to implement this I'm willing to donate a few bucks as well for the effort :-)

erayd:
We will implement this in v3. You don't need to pay us :-).

rbuzatu90:
Hey thanks for such a quick reply! I saw the features planned for v3, but that might take a while and for me this feature would be so handy so get in it sooner would be great.

erayd:
@rbuzatu90 Fair enough - have implemented this in v2 for you (#279).

erayd:
@rbuzatu90 Also done for v3 in browserpass/browserpass-extension#23, if you want to try that version out.

maximbaz:
Merged and released

PR #279, opened 2018-07-24
Allow using left / right arrows to select buttons

erayd:
## What
 * Allow using the left / right arrow keys to navigate between action buttons.

## Why
 * Because @rbuzatu90 specifically requested it for v2 in #267.


rbuzatu90:
Cheeeeers ! :*

PR #300, opened 2018-11-21
Document distribution packages for Debian and Ubuntu.

meskes:


maximbaz:
Really happy to see these packages created! @meskes is that you who maintains them? I'd like to ping you before v3 release to update packaging instructions to make sure smooth transition for users.

meskes:
@maximbaz It's actually done within a team in Debian, but right now I'm the one doing the uploads. Feel free to ping me whenever you are ready.

PR #307, opened 2018-12-29
Add NixOS instructions to README

maximbaz:
Fixes https://github.com/browserpass/browserpass/issues/175#issuecomment-450471165


Infinisil:
See above PR, has instructions for all browsers

PR #308, opened 2018-12-29
Add NixOS instructions

Infinisil:
Refined #307 to work with all browsers

-----

Fixes https://github.com/browserpass/browserpass/issues/175#issuecomment-450471165
Closes #307

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The following OS have a browserpass package that can be installed via package ma
 
 - [Arch Linux](https://aur.archlinux.org/packages/browserpass/)
 - [Debian GNU/Linux](https://tracker.debian.org/pkg/browserpass)
-- [NixOS](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/browserpass/default.nix)
+- [NixOS](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/security/browserpass/default.nix) - make sure to read [these instructions](#configuring-browserpass-on-nixos--for-nix)
 - [Ubuntu](https://launchpad.net/ubuntu/+source/browserpass)
 
 If your OS is not listed above, proceed with the manual installation steps below.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -239,6 +239,54 @@ And finally restart your browser.
...
+  ln -s ~/.nix-profile/lib/mozilla/native-messaging-hosts/com.dannyvankooten.browserpass.json ~/.mozilla/native-messaging-hosts
+# For chrome
+mkdir -p ~/.config/google-chrome/NativeMessagingHosts && \
+  ln -s ~/.nix-profile/etc/chrome-host.json ~/.config/google-chrome/NativeMessagingHosts/com.dannyvankooten.browserpass.json
+# For chromium
+mkdir -p ~/.config/chromium/NativeMessagingHosts && \
+  ln -s ~/.nix-profile/etc/chrome-host.json ~/.config/chromium/NativeMessagingHosts/com.dannyvankooten.browserpass.json
+# For vivaldi
+mkdir -p ~/.config/vivaldi/NativeMessagingHosts && \
+  ln -s ~/.nix-profile/etc/chrome-host.json ~/.config/vivaldi/NativeMessagingHosts/com.dannyvankooten.browserpass.json
```
maximbaz: @Infinisil so Nix package does not put these json files into correct location and every single person has to manually symlink them after installation?

Infinisil: @maximbaz Currently yes. While it is not in the spirit of Nix to require this, it can be very hard to patch browsers to look in another directory for these files. Firefox is the only one I know of that does this, and chrome might be impossible due to it being closed-source.

So on NixOS it is possible to use the `/etc` directory for chrome-based browsers and that's exactly what the `programs.browserpass.enable` option does (see https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/browserpass.nix).

In user-installations of Nix however we don't want to require users write to `/etc`, so using the user-wide paths is perfect.

There is the upcoming project [home-manager](https://github.com/rycee/home-manager) which a lot of people are using already. I might write a home-manager module for browserpass in the future which would remove the need to symlink these files manually for users of it. home-manager or something similar might become part of Nix itself in the future too.

So until this is the case, this simple manual setup is the best we can do. If you wish so, I can also modify the `install.sh` script to work properly on NixOS/Nix.

maximbaz: Thanks for the detailed answer! Let's not update install.sh for now, I'm not exactly sure yet how it will look like with v3, but I'll keep Nix in mind 👍

PR #335, opened 2019-04-07
Prepare notification about v3 release

max-baz:
Trying to keep the message short, @erayd any comments on the wording of the message?

This is how the notification looks for me:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,16 @@
 # Browserpass
 
+**IMPORTANT: this repository is archived and not maintained anymore.**
+
+**Browserpass was rewritten from scratch and split in two repositories:**
+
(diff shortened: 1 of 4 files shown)


erayd (changes requested):
Looks good, but directs users to the browserpass org on GitHub for more info - and there is no obvious additional info there.

Can you add detailed info to the top of the readme file in this PR also?

erayd (changes requested):
Can we also put a FAQ section at the top of the readme? Things like what will happen with OTP, that v3 extension and host both need to be upgraded at the same time, that most distros are packaging the v3 host, etc. I think this would help reduce the risk of someone deciding to fork v2.

erayd (approved):

